PR #57, opened 2015-03-25
Handle exception in requests.response.text.

bsimpson63:
This seems like a bug in the requests package.

The exception occurs here:
https://github.com/kennethreitz/requests/blob/master/requests/models.py#L773

because `encoding` is `None`

fixes the "Bassoons" exception

:eyeglasses: @umbrae @kemitche @dwick


kemitche:
Why not switch from `r.text` to `r.content` and avoid the encoding? (may need to change the log line to be `%r` too)

kemitche:
:nail_care:. If `r.text` is desired/required, consider this a fish.

bsimpson63:
:haircut:

kemitche:
:fish:

umbrae:
:fish:

PR #58, opened 2015-03-26
Gold sidebar ad

MelissaCole:
:eyeglasses: @bsimpson63 
Impose the progress bar and text on top of the background image specified in
the wiki depending on the current gold goal percentage. This image is then
saved and uploaded to adzerk.

Sample dynamic gold wiki: https://gist.github.com/MelissaCole/fe4cebd93d8f0e030021
Can manually on staging and the ad is targeted to /r/subreddit_testing_123 on prod.

Still need to create puppet config to run the cron job
Couldn't find a font library to choose from available fonts on the server like we talked about, so it looks like we might have to push out fonts and specify the path

### DIFF
--- a/reddit_adzerk/adzerk_api.py
+++ b/reddit_adzerk/adzerk_api.py
@@ -54,9 +54,12 @@ class Base(object):
     _fields = FieldSet()
 
     @classmethod
-    def _headers(cls):
-        return {'X-Adzerk-ApiKey': g.secrets['az_selfserve_key'],
+    def _headers(cls, content='urlencoded'):
+        if content == 'urlencoded':
+            return {'X-Adzerk-ApiKey': g.secrets['az_selfserve_key'],
                 'Content-Type': 'application/x-www-form-urlencoded'}
+        else:
+            return {'X-Adzerk-ApiKey': g.secrets['az_selfserve_key']}
 
     def __init__(self, Id, _is_response=False, **attr):
         self.Id = Id
@@ -322,6 +325,14 @@ class Creative(Base):
         items = content.get('items')
         if items:
             return [cls._from_item(item) for item in items]
+
+    @classmethod
+    def upload(cls, Id, image):
+        url = '/'.join([cls._base_url, cls._name, str(Id), 'upload'])
+        response = requests.post(url,
+            headers=cls._headers(content=None), files=image)
+        item = handle_response(response)
+        return cls._from_item(item)
 
     def __repr__(self):
         return '<Creative %s>' % (self.Id)
--- a/reddit_adzerk/dynamicads.py
+++ b/reddit_adzerk/dynamicads.py
@@ -1,0 +1,215 @@
+import ImageDraw, ImageFont
+import os
+from collections import defaultdict
+from pylons import g
+
+from r2.lib.media import url_to_image
+from r2.models import Frontpage
+from r2.models.gold import GoldGoal
+from r2.models.wiki import WikiPageIniItem
+from reddit_adzerk import adzerk_api
+
+default_font_path = "r2/public/static/fonts/"
+default_font = "Georgia.ttf"
+output_path = "r2/public/static/"
+
+class DynamicCreativeWiki(WikiPageIniItem):
+    @classmethod
+    def _get_wiki_config(self):
+        return Frontpage, g.wiki_page_dynamic_gold_ad
+
+    def __init__(self, id, creative_id=None, item_type=None, image_url=None,
+            x_offset=0, x_offset_len=300, y_offset=0, y_offset_len=250,
+            font="Georgia.ttf", font_color="#9A7D2E", font_size=16,
+            align="center", min_percentage=0, max_percentage=100, **kwargs):
+        """reads in config for dynamic gold ad. each creative should have
+        sections with item_types of 'progress_bar' and 'bg_image'.
+        'font_prefs' is optional. all section names should be unique.
+
+        available options for each item_type (*required):
+        [progress_bar]*: creative_id*, item_type*, image_url*,
+            x_offset, y_offset
+        [bg_image]*: creative_id*, item_type*, image_url*,
+            min_percentage, max_percentage, x_offset, y_offset
+        [font_prefs]: creative_id*, item_type*, align,
+            font, font_color, font_size,
+            x_offset, x_offset_len, y_offset, y_offset_len
+        """
+        self.is_enabled = True
+        self.id = id
+        self.creative_id = creative_id
+        self.item_type = item_type
+        self.offset = (int(x_offset), int(y_offset))
+
+        if not creative_id:
+            g.log.error("DynamicCreativeWiki: No creative_id for [%s]" % id)
+        if not item_type or item_type not in ('progress_bar', 'bg_image', 'font_prefs'):
+            g.log.error("DynamicCreativeWiki: Invalid item_type %s for [%s]"
+                % (item_type, self.id))
+
+        if item_type == 'font_prefs':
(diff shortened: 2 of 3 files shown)


reddit_adzerk/dynamicads.py
```diff
@@ -0,0 +1,215 @@
...
+            g.log.error("Creative %s is missing an item" % creative_id)
+            continue
+
+        dc.font_prefs = font_prefs.get(creative_id, None)
+        dc.progress_bar = progress_bars[creative_id]
+        dc.image_collection = bg_images[creative_id]
+        dc.build_creative()
+
+        # upload the image file to adzerk
+        image = {'image': open(dc.output_location, 'rb')}
```
JordanMilne: Why do we save and then reopen the same image? Is the save to file step necessary?

JordanMilne: If it's just so that `requests` gets a filename for the upload, instead of a `file` object you can pass in a tuple of `(filename, data)`, see http://docs.python-requests.org/en/latest/user/quickstart/#post-a-multipart-encoded-file

reddit_adzerk/adzerk_api.py
```diff
@@ -54,9 +54,12 @@ class Base(object):
     _fields = FieldSet()
 
     @classmethod
-    def _headers(cls):
-        return {'X-Adzerk-ApiKey': g.secrets['az_selfserve_key'],
+    def _headers(cls, content='urlencoded'):
```
bsimpson63: looks like `content` is never used

reddit_adzerk/adzerk_api.py
```diff
@@ -323,6 +326,14 @@ def list(cls, AdvertiserId):
         if items:
             return [cls._from_item(item) for item in items]
 
+    @classmethod
+    def upload(cls, Id, image):
+        url = '/'.join([cls._base_url, cls._name, str(Id), 'upload'])
+        response = requests.post(url,
+            headers=cls._headers(content=None), files=image)
```
bsimpson63: What does adzerk assume is the content type if not specified? What happens when you use 'application/x-www-form-urlencoded' for this post?

Might make sense to have 'application/x-www-form-urlencoded' be the default and then you could override it if needed.

reddit_adzerk/dynamicads.py
```diff
@@ -0,0 +1,215 @@
...
+            min_percentage, max_percentage, x_offset, y_offset
+        [font_prefs]: creative_id*, item_type*, align,
+            font, font_color, font_size,
+            x_offset, x_offset_len, y_offset, y_offset_len
+        """
+        self.is_enabled = True
+        self.id = id
+        self.creative_id = creative_id
+        self.item_type = item_type
+        self.offset = (int(x_offset), int(y_offset))
```
bsimpson63: It might be nice to convert all the ints to ints early in this method.

Also looks like `WikiPageIniItem` only supports `_bool_values` for coercing type. Possibly worth adding support for ints, floats, etc?

reddit_adzerk/dynamicads.py
```diff
@@ -0,0 +1,215 @@
...
+
+        if not creative_id:
+            g.log.error("DynamicCreativeWiki: No creative_id for [%s]" % id)
```
bsimpson63: Should there be an exception here and below, or it's ok to partially instantiate the object (that will be detected and rejected later)?

reddit_adzerk/dynamicads.py
```diff
@@ -0,0 +1,215 @@
...
+        self.progress_bar = None
+        self.font_prefs = None
+        self.image_collection = []
+
+        self.background_item = None
+        self.background_image = None
+
+        self.output_path = output_path
+        self.output_name = "goldvertisement_" + str(creative_id) + ".png"
+        self.output_location = self.output_path + self.output_name
```
bsimpson63: maybe use `os.path` here

reddit_adzerk/dynamicads.py
```diff
@@ -0,0 +1,215 @@
...
+
+    def set_background(self):
+        '''choose the background image based on the current gold goal percentage'''
+        for image in self.image_collection:
+            if self.percent >= image.min_percentage:
+                # if max_percentage doesn't exist or is 100, use for percent >= 100
```
bsimpson63: might be clearer to special case `self.percent >= 100` and otherwise check `if image.min_percentage <= self.percent <= image.max_percentage`

reddit_adzerk/dynamicads.py
```diff
@@ -0,0 +1,215 @@
...
+        self.background_image = url_to_image(str(self.background_item.image_url))
+
+    def build_creative(self):
+        '''choose the background image based on the percentage, impose the progress_bar
+        on top of it. add text on top if a font_prefs section exists.
+        save the image to a file which can then be sent to adzerk
+        '''
+        self.set_background()
+        progress_bar = url_to_image(self.progress_bar.image_url)
+        self.draw_progress_bar(progress_bar, self.progress_bar.offset, self.percent/100.0)
```
bsimpson63: looks like you're not ding anything with the return from this

reddit_adzerk/dynamicads.py
```diff
@@ -0,0 +1,215 @@
...
+        self.background_image.paste(progress_bar, self.background_item.offset, progress_bar)
+
+        if self.font_prefs:
+            self.draw_text(self.background_image)
```
bsimpson63: here too

reddit_adzerk/dynamicads.py
```diff
@@ -0,0 +1,215 @@
...
+
+        if not os.path.exists(self.output_path):
+            os.makedirs(self.output_path)
+        self.background_image.save(self.output_location)
```
bsimpson63: Is there anything we need to worry about with creating directories and files? What if we don't have permissions? Anything else @JordanMilne?

reddit_adzerk/dynamicads.py
```diff
@@ -0,0 +1,215 @@
...
+
+    def draw_progress_bar(self, image, offset, percent):
+        '''draw a rectangle on the progress_bar outside of the current gold goal %
+        and fill it with white'''
+        draw = ImageDraw.Draw(image)
+        draw.rectangle(
+            [(offset[0]+(image.size[0]*percent), offset[1]),
+                (offset[0]+image.size[0], offset[1]+image.size[1])],
+            fill=(255, 255, 255))
```
bsimpson63: does this mean that the progress bar color is not configurable?

reddit_adzerk/dynamicads.py
```diff
@@ -0,0 +1,215 @@
...
+
+        available options for each item_type (*required):
+        [progress_bar]*: creative_id*, item_type*, image_url*,
+            x_offset, y_offset
+        [bg_image]*: creative_id*, item_type*, image_url*,
+            min_percentage, max_percentage, x_offset, y_offset
+        [font_prefs]: creative_id*, item_type*, align,
+            font, font_color, font_size,
+            x_offset, x_offset_len, y_offset, y_offset_len
+        """
```
bsimpson63: Can you post an example config?

bsimpson63: It looks like you have multiple sections per creative_id (progress_bar, bg_image, font_prefs). Why did you do this rather than have a single section per creative_id that defines all the options?

bsimpson63:
:nail_care:

zeantsoi:
@MelissaCole this looks pretty stale. Close out?

MelissaCole:
Indeed! Closing

PR #59, opened 2015-03-26
Use platform and mobile OS targeting.

madbook:
for platform targeting, pass along the appropriate site id.  for os targeting, we need to add a custom targeting 'zerkel' query to the flight and pass the user agent along when making the request.  I have _not_ tested this on stagin yet (going to try to get to that today).

:eyeglasses: @bsimpson63

### DIFF
--- a/reddit_adzerk/adzerkpromote.py
+++ b/reddit_adzerk/adzerkpromote.py
@@ -236,11 +236,11 @@ def update_flight(link, campaign, az_campaign):
         })
 
     if campaign.mobile_os:
-        deviceQueries = ['($device.os = "%s")' % os
+        deviceQueries = ['($device.os contains "%s")' % os
                          for os in campaign.mobile_os]
 
         if campaign.platform == "all":
-            deviceQueries.append('($device.formFactor = "desktop")')
+            deviceQueries.append('($device.formFactor contains "desktop")')
         
         customTargeting = ' or '.join(deviceQueries)
         d.update({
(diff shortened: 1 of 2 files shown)


bsimpson63:
This depends on https://github.com/reddit/reddit-public/pull/3341/ right?

bsimpson63:
Looks good, let me know when you've tested this and I'll give it the stamp of approval.

madbook:
had a couple of errors to fix up.  the small amount of testing I did yesterday looked good, but I want to run through the various configurations to make sure they're all working as expected.  I'll haircut when ready

madbook:
Ok, I think this is all good now. Just those couple of small changes in the last 3 commits. :haircut:

reddit_adzerk/adzerkpromote.py
```diff
@@ -234,6 +235,36 @@ def update_flight(link, campaign, az_campaign):
...
+            })
+        elif campaign.platform == 'mobile':
+            siteZones.append({
+                'SiteId': g.az_selfserve_mobile_web_site_id,
+                'IsExclude': False,
+            })
+
+        if len(siteZones):
+            d.update({
+                'SiteZoneTargeting': siteZones
```
bsimpson63: just checking: all existing ads were not specifying 'SiteZoneTargeting", so they are essentially targeted to all sitezones?

bsimpson63: ah, and I guess that's the same reason we only add it here if `platform != 'all'`

bsimpson63:
:fish:

reddit_adzerk/adzerkpromote.py
madbook: yup!

madbook:
4ede381bfebece099ec88e447505243684747d45

PR #60, opened 2015-04-13
Rename Adzerk API response timer for consistency with other timers.

spladug:
All of the providers are moving under a single tree for consistency and
cleanliness.

:eyeglasses: @bsimpson63


bsimpson63:
:fish:

PR #61, opened 2015-04-20
POST_request_promo: use `responsive` to wrap response.

bsimpson63:
This ensures we send proper json for api requests.

:eyeglasses: @kemitche @ajacksified


ajacksified:
:runner: I defer to @kemitche

kemitche:
:fish:, don't see why not

PR #67, opened 2015-06-01
Frontpage stats

dwick:
:eyeglasses: @zeantsoi @rram 

instrumented some more things in statsd to help with the debugging


zeantsoi:
Stats make sense. Would it be better to abstract into a separate function so that it could be invoked as a one-liner? Something like:

    # function definition
    def frontpage_stat(stat, targeting_frontpage)
        if targeting_frontpage:
            g.stats.simple_event("adzerk.frontpage_target.request.%s" % stat)

    # example function invocation
    frontpage_stat('refused', targeting_frontpage)

zeantsoi:
:nail_care:

rram:
Sorry, I don't have time for this. I've nominated @bsimpson63 in my place :runner:

bsimpson63:
:nail_care:

dwick:
@bsimpson63 since I'm trying to figure out what's going on with frontpage under delivery as much info as possible is helpful

:haircut:

bsimpson63:
no_decisions/no_promo is still redundant right?

dwick:
I think it's useful to not have to subtract all the connection errors to figure out if it was an error or adzerk actually came back with nothing. (coming back with nothing is what i'm most interested in)

dwick:
:haircut:

bsimpson63:
:fish:

PR #68, opened 2015-07-14
Handle extra param for identifying user for frequency capping

zeantsoi:
:eyeglasses: @dwick @curioussavage 

Per the [Adzerk docs](http://help.adzerk.com/hc/en-us/articles/202533809), we need to add a user key in order to use UserDB. This will enable a frequency cap to adhered to when requesting ads.

When a user is logged out, we grab the `loid` cookie value and pass that as the unique identifier. When the user is logged in, we pass the user's ID. If the campaign requested is frequency capped, it will accept the user key and distribute accordingly.


curioussavage:
:fish:

reddit_adzerk/adzerkpromote.py
```diff
@@ -602,8 +607,9 @@ class AdzerkApiController(api.ApiController):
     @validate(
         srnames=VPrintable("srnames", max_length=2100),
         is_mobile_web=VBoolean('is_mobile_web'),
+        loid=nop('loid'),
     )
-    def POST_request_promo(self, srnames, is_mobile_web):
+    def POST_request_promo(self, srnames, is_mobile_web, loid):
```
dwick: should this param be user_id?  it's not always the logged out id

reddit_adzerk/adzerkpromote.py
```diff
@@ -645,3 +652,11 @@ def POST_request_promo(self, srnames, is_mobile_web):
             return responsive(w.render(), space_compress=True)
         else:
             g.stats.simple_event('adzerk.request.skip_promo')
+
+    def get_uid(self, loid):
+        if loid:
+            return loid
+        elif c.user_is_loggedin:
```
dwick: this should probably be the first branch in case `loid` is passed accidentally for some reason

reddit_adzerk/adzerkpromote.py
```diff
@@ -602,8 +607,9 @@ class AdzerkApiController(api.ApiController):
     @validate(
         srnames=VPrintable("srnames", max_length=2100),
         is_mobile_web=VBoolean('is_mobile_web'),
+        loid=nop('loid'),
```
dwick: should we get more specific on the validator here?  not sure what `loid`s look like. guid??

also, defaulting it to `None` would make it more clear that it's optional to me

dwick:
:nail_care: small stuff

dwick:
are we missing setting of the actual frequency capping params when creating flights/whatever in adzerk?

reddit_adzerk/adzerkpromote.py
zeantsoi: Talked this over with Chris Stephens, who implemented `loid`. It may not be of much use to try to validate against this since:

1. It seems like it's still relatively new, somewhat inconsistent, and possibly in a state of flux
2. The only thing we could validate is that it's alphanumeric and isn't a super huge blog of garbage.

It's the first thing that worries me about validating at this point... I think this is something we can revisit if it ends up causing any problems.

I've updated so the default is set to `None`.

zeantsoi:
:haircut: @dwick

dwick:
:fish: please update the commit message.  "Add frequency capping" seems reasonable to me

zeantsoi:
5349b3c3c73237d47095e2bd3e35ab68e733b16d

PR #69, opened 2015-07-18
Decorate Adzerk request to enable identification via OAuth

zeantsoi:
:eyeglasses: @dwick 

I still can't get my local instance to spin up, so this is untested. If I'm missing something, please let me know.


dwick:
:fish:, you could test on staging

PR #70, opened 2015-07-21
Assign uid to pass to Adzerk reqeust

zeantsoi:
:eyeglasses: @dwick 

Pretty sure that the issue with 6c844a2afa89b92619c05d43a6e45af7f2b1b315 is that the `get_uid` function is not decorated with the `allow_oauth2_access` wrapper, so it doesn't have access to the `c.user` context.

Rather than passing more vars into `get_uid` or adding `ValueError` exception handling, I figure it's probably more concise to move the logic of assigning `uid` back into the controller action itself.

I still haven't been able to test this either locally or on staging, but once I'm able to do so, it should be a pretty straightforward to determine whether this entirely resolves the issue.


dwick:
i think we can close this one

PR #71, opened 2015-07-21
Pass id36 to Adzerk rather than id

zeantsoi:
:eyeglasses: @dwick 

There may be an issue with passing an `int` type to the Adzerk request. This will pass a `str` type, which has proven to work in the case of logged out users via the `loid`.


dwick:
:fish:

PR #73, opened 2015-07-30
Click tracking

dwick:

### DIFF
--- a/reddit_adzerk/adzerkpromote.py
+++ b/reddit_adzerk/adzerkpromote.py
@@ -643,13 +643,14 @@ class AdzerkApiController(api.ApiController):
             return responsive(response.body)
 
         res_by_campaign = {r.campaign: r for r in response}
+        adserver_click_urls = {r.campaign: r.click_url for r in response}
         tuples = [promote.PromoTuple(r.link, 1., r.campaign) for r in response]
         builder = CampaignBuilder(tuples, wrap=default_thing_wrapper(),
                                   keep_fn=promote.promo_keep_fn,
                                   num=1,
                                   skip=True)
         listing = LinkListing(builder, nextprev=False).listing()
-        promote.add_trackers(listing.things, c.site)
+        promote.add_trackers(listing.things, c.site, adserver_click_urls=adserver_click_urls)
         if listing.things:
             g.stats.simple_event('adzerk.request.valid_promo')
             w = listing.things[0]


dwick:
https://github.com/reddit/reddit-public/pull/3980

dwick:
@zeantsoi i need to update all the ads with their url, @bsimpson63 said you had to queue live ads manually last week?

bsimpson63:
I think I remember the issue with this and why it was removed.

The adzerk click redirect url includes (base64 encoded) details about the initial request for an ad, which for us includes the names of subreddits that a user is subscribed to. So the destination server (that handles the ad url) would get the http referer and this was deemed a privacy issue.

This might not be a problem now if:
* adzerk has changed how it forms the click redirect url (I have no idea if this is the case)
* we force all traffic to go through adzerk https the referer won't be sent (I don't know under which conditions this is true)
* this is just an acceptable information leak

zeantsoi:
The problem with queuing was related to a bug that was resolved with [this commit](https://github.com/reddit/reddit-public/commit/497875e4589f1e885dbdb54eb1fdb139d8e52e39) in core.

I'm actually  unfamiliar with this particular part of the code. I have been working, however, with the requests made in `tracking.py`. In this request, we send a base64 encoded parameter that contains user-related information, but not anything _specific to the user_, such a subscribed subreddits. [Here's the code](https://github.com/reddit/reddit-public/blob/master/r2/r2/lib/tracking.py#L152-L160).

Is there actually a privacy related issue with this?

reddit_adzerk/adzerkpromote.py
```diff
@@ -643,13 +643,14 @@ def POST_request_promo(self, srnames, is_mobile_web, loid):
...
         res_by_campaign = {r.campaign: r for r in response}
+        adserver_click_urls = {r.campaign: r.click_url for r in response}
         tuples = [promote.PromoTuple(r.link, 1., r.campaign) for r in response]
         builder = CampaignBuilder(tuples, wrap=default_thing_wrapper(),
                                   keep_fn=promote.promo_keep_fn,
                                   num=1,
                                   skip=True)
         listing = LinkListing(builder, nextprev=False).listing()
-        promote.add_trackers(listing.things, c.site)
+        promote.add_trackers(listing.things, c.site, adserver_click_urls=adserver_click_urls)
```
zeantsoi: `add_trackers` [accepts just two args](https://github.com/reddit/reddit-public/blob/master/r2/r2/lib/promote.py#L179) right now.

dwick: https://github.com/reddit/reddit-public/pull/3980

bsimpson63:
@zeantsoi I think the existing stuff in `_get_encrypted_user_slug()` is fine, but it is a bit more questionable to include a list of the user's subscribed subreddits (like adzerk does or at least used to do).

dwick:
@bsimpson63 

screen shot 2015-07-30 at 2 48 17 pm

bsimpson63:
@dwick what point in the request is that?

dwick:
that's the adzerk redirect after our redirect

dwick:
screen shot 2015-07-30 at 3 06 38 pm

dwick:
:haircut:

bsimpson63:
:fish:

dwick:
Going to deploy this in 2 parts, 1 to update the urls and then switch the redirects when I'm happy with what I see on staging. //cc @bsimpson63

dwick:
https://github.com/reddit/reddit-plugin-adzerk/commit/cda959de63d7bb0700bc3a7b0f63b7e6ab20c699

dwick:
this is up on my staging https://dwick.staging.wireddit.com/

dwick:
any ideas on how to get this info here: https://github.com/reddit/reddit-public/blob/2827255b094d00823a15a46b18a7f801ca4db434/r2/r2/controllers/listingcontroller.py#L644

PR #74, opened 2015-08-04
Revert "Temp: stop passing url"

dwick:
Intermittent adzerk issue caused a scare.  This should be fine now.

:eyeglasses: @bsimpson63


bsimpson63:
:fish:

PR #78, opened 2015-08-07
Add Zerkel queries for device and version

zeantsoi:
:eyeglasses: @dwick 

I was able to test this using a fake campaign object here. You can load it into shell using [this script](https://gist.github.com/zeantsoi/c05debd8f6d80983c657).

For a campaign with the following attributes:

- `mobile_os`: `['iOS', 'Android']`
- `ios_devices`: `['iPhone', 'iPad']`
- `android_devices`: `['tablet']`
- `ios_version_range`: `['1.1', '']`
- `android_version_range`: `['4.4', '4.4']`

The following query is generated:

    ($device.formFactor contains "desktop") or ($device.os = "iOS" and ($devices.modelName like "iPhone" or $devices.modelName like "iPad") and (($device.osVersion.major >= 2) or ($device.osVersion.major = 1 and $device.osVersion.minor >= 1))) or ($device.os = "Android" and ($devices.formFactor like "tablet") and (($device.osVersion.major = 4 and $device.osVersion.minor = 4)))


reddit_adzerk/adzerk_utils.py
```diff
@@ -0,0 +1,61 @@
...
+    if not version_range:
+        return
+
+    def join_queries(operator, *args):
+        delimiter = ' %s ' % operator
+        return delimiter.join(args)
+
+    lower_major, lower_minor = [int(bound) for bound in version_range[0].split('.')]
+
+    lower_template = '($device.osVersion.major = %s and $device.osVersion.minor >= %s)'
```
JordanMilne: Can these unquoted `%s`s be `%d`s?

reddit_adzerk/adzerkpromote.py
```diff
@@ -196,6 +197,26 @@ def update_creative(link, campaign):
     return az_creative
 
 
+def get_mobile_targeting_query(os_str, lookup_str, campaign_mobile_os,
+                               devices, versions, get_version_query):
+    if os_str in campaign_mobile_os:
+        queries = []
+
+        os_query = '$device.os = "%s"' % os_str
```
JordanMilne: Looks like this might let you escape the query if you specify an OS like `foo" <stuff>`. Does Zerkel support either single-quoted or double-quoted strings? If so, you can do something like `'$device.os = %r' % os_str` to template in properly escaped string. Otherwise you'll have to do something like:

```python
def str_to_c_str(val):
    return '\"%s\"' % (val.encode("string_escape").replace('"', r'\"'))
```

which'll give you

```python
>>> print str_to_c_str("foo\" <stuff>")
"foo\" <stuff>"
```

reddit_adzerk/adzerkpromote.py
```diff
@@ -196,6 +197,26 @@ def update_creative(link, campaign):
     return az_creative
 
 
+def get_mobile_targeting_query(os_str, lookup_str, campaign_mobile_os,
```
dwick: this could probably go in adzerk utils too

reddit_adzerk/adzerkpromote.py
```diff
@@ -196,6 +197,26 @@ def update_creative(link, campaign):
...
+                               devices, versions, get_version_query):
```
dwick: does `get_version_query` need to be passed in?  will this ever accept a different function?

reddit_adzerk/adzerkpromote.py
zeantsoi: To my knowledge, Zerkel supports double-quoted string _only_. This function should be invoked only [here](https://github.com/reddit/reddit-plugin-adzerk/pull/78/files#diff-8ff7bdcffc2cdf19a06e2d539306378aR275) and [here](https://github.com/reddit/reddit-plugin-adzerk/pull/78/files#diff-8ff7bdcffc2cdf19a06e2d539306378aR281); both instances invoke predefined, properly-formatted, static strings.

Overall, I think that string escaping (in this case) is a bit overkill. I have, however, added a preceding `_` to the function name to denote the private nature of its use.

reddit_adzerk/adzerkpromote.py
```diff
@@ -247,16 +268,29 @@ def update_flight(link, campaign, az_campaign):
             'RateType': 1, # 1: Flat
         })
 
+    # Zerkel queries here
     if campaign.mobile_os:
-        deviceQueries = ['($device.os contains "%s")' % os
-                         for os in campaign.mobile_os]
+        queries_list = []
+
+        queries_list.append(get_mobile_targeting_query('iOS', 'modelName',
```
dwick: nit: probably best to use named parameters for a function with so many arguments

reddit_adzerk/adzerkpromote.py
```diff
@@ -247,16 +268,29 @@ def update_flight(link, campaign, az_campaign):
...
+                                                       campaign.mobile_os,
+                                                       campaign.ios_devices,
+                                                       campaign.ios_version_range,
+                                                       get_version_query))
+
+        queries_list.append(get_mobile_targeting_query('Android', 'formFactor',
```
dwick: same as above

dwick:
This seems like a prime candidate for unit tests, any chance you'd want to add a couple?  I know you have the code basically already from that gist you sent around yesterday.  Overall looks good though. :+1: 

:nail_care:

reddit_adzerk/adzerkpromote.py
zeantsoi: I had some bad scoping in testing. It does *not* need to be passed in, but I've moved it into the `adzerk_utils.py` file anyways.

zeantsoi:
:haircut: @dwick 

Tests sound great, and I think this is probably a pretty slam dunk candidate. I'll work on them, but they may need to go out post-production.

I'll also work on integrating a few small suggestions that came through from Adzerk this morning. I'll have them in code ahead of a planned Monday morning deploy.

reddit_adzerk/adzerk_utils.py
```diff
@@ -0,0 +1,86 @@
...
+        if devices and versions:
+            device_queries = ['$devices.%s like "%s"' % (lookup_str, device)
+                              for device in devices]
+            device_query = '(%s)' % ' or '.join(device_queries)
+            version_query = '(%s)' % get_version_query(versions)
+
+            queries.append(device_query)
+            queries.append(version_query)
+
+    return '(%s)' % (' and ').join(queries)
```
dwick: could use `join_queries` here

zeantsoi: Done.

dwick:
lgtm, haircut again when you make adzerk's suggested changes :nail_care:

zeantsoi:
:haircut: [minor typo](https://github.com/zeantsoi/reddit-plugin-adzerk/commit/7d2a63807d680793908de83b74796f53988c13ec). That's all from Adzerk...

zeantsoi:
:haircut:

dwick:
:fish:

zeantsoi:
73dbe33a59cfbc019e8115dfc9b93b45eecc685a

PR #80, opened 2015-08-12
Return None if OS is not present

zeantsoi:
:eyeglasses: @kjoconnor @bsimpson63 @Deimos


bsimpson63:
:fish:

zeantsoi:
:eyeglasses: @bsimpson63 

`adzerkpromote.py` is the only file that invokes `get_mobile_targeting_query`.

PR #82, opened 2015-08-13
Make operators upper and swap LIKE for CONTAINS

zeantsoi:

According to Adzerk, the instances where they told us to use `LIKE` should actually be `CONTAINS`. I've tested it in the Adzerk UI and `CONTAINS` seems to work; I've made the change.

Also, it looks like the [Adzerk parser](https://github.com/adzerk/zerkel/blob/master/src/zerkel-parser.js#L638) accepts case-insensitive logical operators. I've switched all our operators to uppercase to better conform with general SQL conventions.


reddit_adzerk/adzerk_utils.py
```diff
@@ -75,12 +75,12 @@ def get_mobile_targeting_query(os_str='',
     queries.append(os_query)
 
     if devices and versions:
-        device_queries = ['$device.%s like "%s"' % (lookup_str, device)
+        device_queries = ['$device.%s CONTAINS "%s"' % (lookup_str, device)
```
bsimpson63: can this be a separate commit?

zeantsoi:
:haircut:

bsimpson63:
:fish:

PR #83, opened 2015-08-13
Swap LIKE for CONTAINS

zeantsoi:

Per James from Adzerk:

> The first is that there is a problem with the Zerkel, sorry we didn't catch it earlier, you need to use CONTAINS not "like". (you could use like with * around the text, but contains is going to be more efficient for this use)

Change `LIKE` to `CONTAINS`.


bsimpson63:
:fish:

dwick:
:fish:

zeantsoi:
120caa90897524010e476ccfbf3f1949f038946d

PR #86, opened 2015-08-21
Remove the relationship between `PromoCampaign`s and creatives.

dwick:
:eyeglasses: @bsimpson63 @zeantsoi 

I'm a little fuzzy on the static properties of `tdb_cassandra.View` so i'd definitely look at that.


reddit_adzerk/models/flight.py
```diff
@@ -0,0 +1,44 @@
...
+
+    @classmethod
+    def _rowkey(cls, campaign):
+        return campaign.adzerk_flight_id
+
+    @classmethod
+    def add(cls, campaign):
+        rowkey = cls._rowkey(campaign)
+        cls._set_values(rowkey, {campaign._fullname: ""})
+        return text
```
bsimpson63: what's `text`?

reddit_adzerk/models/flight.py
```diff
@@ -0,0 +1,44 @@
+from r2.lib.db import tdb_cassandra
+from r2.models import PromoCampaign
+
+
+def _lookup(flight_id):
```
bsimpson63: don't feel strongly, but this could be a classmethod of `PromoCampaignByFlightId`

reddit_adzerk/adzerkpromote.py
```diff
@@ -617,10 +618,14 @@ def adzerk_request(keywords, uid, num_placements=1, timeout=1.5,
         if decision['campaignId'] in g.adserver_campaign_ids:
             return AdserverResponse(decision['contents'][0]['body'])
 
+        adzerk_flight_id = decision['flightId']
         imp_pixel = decision['impressionUrl']
         click_url = decision['clickUrl']
+
+        # lookup campaign for flight
+        campaign = PromoCampaignByFlightId.get(adzerk_flight_id)
```
bsimpson63: What happens if the campaign is not found?

bsimpson63: the comment seems unnecessary

bsimpson63:
:nail_care:

reddit_adzerk/models/flight.py
```diff
@@ -0,0 +1,44 @@
...
+        return text
+
+    @classmethod
+    def get(cls, flight_id):
+        try:
+            row = cls._byID(flight_id)
+        except tdb_cassandra.NotFound:
+            campaign = _lookup(flight_id)
+
+            if campaign:
```
zeantsoi: Could there be an edge case where the lookup fails to return a campaign, and if so, should we handle that?

reddit_adzerk/models/flight.py
```diff
@@ -0,0 +1,44 @@
...
+                cls.add(campaign)
+
+                return campaign._fullname
+
+        values = row._values()
```
zeantsoi: Since this logic presumes that `row` is set, maybe encapsulate it in the `else` condition of a `try-except-else` handler?

reddit_adzerk/models/flight.py
dwick: since this wouldn't actually use `cls` i see no reason to add it to the class.

dwick:
:haircut:

reddit_adzerk/models/flight.py
zeantsoi: Since this is a private method being used just once, does it make sense to put it directly into the invoking `get` method? Otherwise, could it be a class method on `PromoCampaign`?

reddit_adzerk/adzerkpromote.py
```diff
@@ -617,10 +618,16 @@ def adzerk_request(keywords, uid, num_placements=1, timeout=1.5,
         if decision['campaignId'] in g.adserver_campaign_ids:
             return AdserverResponse(decision['contents'][0]['body'])
 
+        adzerk_flight_id = decision['flightId']
         imp_pixel = decision['impressionUrl']
         click_url = decision['clickUrl']
+
+        campaign = PromoCampaignByFlightId.get(adzerk_flight_id)
+
+        if not campaign:
```
bsimpson63: Maybe add `g.stats.simple_event` and a `g.log.error` here.

Might be over-engineering things, but you could also do some cleanup for this case by trying to figure out which PromoCampaign should be associated with the Flight.

dwick: this shouldn't happen ever. i can definitely add some logging just in case i'm wrong.

> trying to figure out which PromoCampaign should be associated with the Flight

suggestion on how to do so?

reddit_adzerk/models/flight.py
dwick: i think moving this to `_query_one` on `Thing` would be useful, @bsimpson63 would you accept a PR for that?  otherwise i'm inclined to just leave it as a private method.

reddit_adzerk/adzerkpromote.py
```diff
@@ -617,10 +618,19 @@ def adzerk_request(keywords, uid, num_placements=1, timeout=1.5,
...
+            g.stats.simple_event('adzerk.request.missing_campaign')
+            g.log.error('adzerk_request: couldn\'t find campaign for flight (%s)',
+                adzerk_flight_id)
+            continue
```
bsimpson63: do you need to just return here?

reddit_adzerk/models/flight.py
bsimpson63: I think just leave as is.

reddit_adzerk/adzerkpromote.py
dwick: spoke with brian offline. no change here, just keep an eye on the logs

reddit_adzerk/adzerkpromote.py
dwick: there could be more than 1 response to process, so no

bsimpson63:
:fish:

dwick:
:eyeglasses: @spladug 

:haircut: @bsimpson63

reddit_adzerk/lib/cache.py
```diff
@@ -0,0 +1,39 @@
...
+
+    l = list(q)
+    if l:
+        return l[0]
+
+
+class PromoCampaignByFlightIdCache():
+    @classmethod
+    def _cachekey(cls, flight_id):
+        return "promo.flight.%s" % str(flight_id)
```
rram: aww yeah structured key names :+1:

spladug: What does a flight_id look like? Can we trust it, or is it coming from external? I'm concerned about passing it verbatim into a cache key without checking it.  If it's a number, doing `%d` would be sufficient for me.

reddit_adzerk/lib/cache.py
```diff
@@ -0,0 +1,39 @@
...
+
+        if not fullname:
+            campaign = _lookup(flight_id)
+
+            if campaign:
+                cls.add(campaign)
+
+                return campaign._fullname
+            else:
+                return None
```
spladug: This negative-result case isn't cached at all. Is it likely to ever happen?

reddit_adzerk/lib/cache.py
```diff
@@ -0,0 +1,39 @@
+from pylons import g
+
+from r2.models import PromoCampaign
+
+
+def _lookup(flight_id):
```
spladug: tiny nit: lookup is a noun, look_up would be better as a function name since it's more verby

reddit_adzerk/lib/cache.py
```diff
@@ -0,0 +1,39 @@
...
+    q = PromoCampaign._query(PromoCampaign.c.adzerk_flight_id == flight_id)
+    q._limit = 1
+
+    l = list(q)
+    if l:
+        return l[0]
+
+
+class PromoCampaignByFlightIdCache():
```
spladug: This class isn't really necessary.  Could just look something like

```python
_KEY_FORMAT = "promo.flight.%s"

def cache_campaign_by_flight(campaign)
    cache_key = _KEY_FORMAT % campaign.adzerk_flight_id
    g.cache.set(cache_key, campaign._fullname, time=60*60*24)

def fetch_campaign_for_flight(flight_id):
    campaign_fullname = g.cache.get(_KEY_FORMAT % flight_id)

    if not campaign_fullname:
        query = PromoCampaign._query(
            PromoCampaign.c.adzerk_flight_id == flight_id,
        )
        query._limit = 1
        campaigns = list(query)
        if not campaigns:
            return None
        campaign = campaigns[0]
        cache_campaign_by_flight(campaign)
        g.cache.set(cache_key, campaign_fullname, time=60*60*24)

    return campaign_fullname
```

(untested)

but I don't feel strongly about this.

spladug:
:nail_care: just a couple of questions. nice!

reddit_adzerk/lib/cache.py
dwick: this would be an orphaned flight.  i want to say it won't happen, but i also want to be sure we handle it so we don't hammer the database with the same bad flight lookup

reddit_adzerk/lib/cache.py
dwick: kept the same public interface when i replaced the CF with memcached.  i'm kind of partial to the class i guess... is there any performance difference  between static classes and modules?

dwick:
:haircut:

reddit_adzerk/lib/cache.py
spladug: no, performance shouldn't matter here anyway (that'd just be microoptimizing CPU without benchmarking to show we should care).  but it's unnecessary code that makes you have to jump around the file to understand what's going on rather than just reading linearly through a function.

spladug:
:fish: classless would still be more readable imo, but this looks fine

reddit_adzerk/adzerk_api.py
```diff
@@ -209,7 +209,20 @@ class Advertiser(Base):
...
+        items = content.get('items')
+
+        return items
+
+    @classmethod
+    def find_one(cls, Title):
+        results = cls.search(Title)
+
+        if results:
+            return results[0]
```
bsimpson63: `find_one` and `search` are never used

reddit_adzerk/adzerkpromote.py
```diff
@@ -397,13 +437,13 @@ def create_cfmap(link, campaign, az_campaign, az_creative, az_flight):
...
-    if getattr(campaign, 'adzerk_cfmap_id', None) is not None:
-        raise AttributeError('%s has existing adzerk_cfmap_id' % campaign)
+    if getattr(campaign, 'az_cfmap_id', None) is not None:
+        raise AttributeError('%s has existing az_cfmap_id' % campaign)
 
     d = {
         'SizeOverride': False,
         'CampaignId': az_campaign.Id,
-        'PublisherAccountId': g.az_selfserve_advertiser_id,
+        'PublisherAccountId': g.az_selfserve_publisher_id,
```
dwick: lost this comment in a rebase @bsimpson63:
> why did this change?

advertisers and publishers are different things in adzerk. we are actually using advertiser accounts for reddit users now so i figured we should name this properly at least.

bsimpson63: Can you break this commit up into the distinct ini changes?

reddit_adzerk/adzerk_api.py
dwick: yeah, i was going to use these to lookup the advertisers by name before i realized caching on the account was the better way to go.  i can remove this commit, but they do work so could go either way for me.

reddit_adzerk/adzerkpromote.py
dwick: done

reddit_adzerk/adzerk_api.py
bsimpson63: I vote for remove

reddit_adzerk/adzerkpromote.py
```diff
@@ -135,12 +147,13 @@ def update_campaign(link):
             log_text = None
     else:
         d.update({
+            'AdvertiserId': az_advertiser.Id,
```
bsimpson63: So when we're just updating an existing campaign it's not expected that the az_advertiser will be passed in? That logic seems a bit loose.

How about just doing something like:

```
if az_advertiser:
    d["AdvertiserId"] = az_advertiser.Id
```

on line 139 after `d` is initially created

dwick:
related config: https://github.com/reddit/reddit-private/pull/1012

Migration plan:

1. All `Temp:` commits are only for use staging during the migration
2. Only scheduled campaigns will be pushed to the new adzerk instance (others will move over when payed for/approved):
```
import promote

errored = []
scheduled = promote.get_scheduled_promos()

for campaign, link in scheduled:
  try:
    _update_adzerk(link=link, campaign=campaign)
  except Exception as e:
    print "error occured: \n\t%s" % str(e)
    errored.append((link, campaign, e))
```
3. Temp commits will be rebased off and changes pushed to production

bsimpson63:
re migration:

What's the post data migration plan? What changes happen in production to start serving ads out of the new account?

reddit_adzerk/adzerkpromote.py
dwick: that's what i had before!

dwick:
Just he config pushes along with updating the job servers

reddit_adzerk/adzerkpromote.py
bsimpson63: sorry!

bsimpson63:
:fish:

dwick:
merged

PR #87, opened 2015-08-27
Use SSL on all adzerk requests

dwick:
:eyeglasses: @bsimpson63


bsimpson63:
:fish:

dwick:
note: need to rotate the key when we push this out

PR #91, opened 2015-09-03
Click tracking

dwick:

related: https://github.com/reddit/reddit-public/pull/4130


reddit_adzerk/adzerkpromote.py
```diff
@@ -202,6 +209,17 @@ def update_creative(link, az_advertiser):
     return az_creative
 
 
+def update_trackers(campaign, az_cfmap):
+    if (not getattr(campaign, "adserver_click_url", False) or
+        not getattr(campaign, "adserver_impression_url", False)):
+
+        pixels = az_cfmap._get_pixel()
+
+        campaign.adserver_click_url = pixels['StaticClickUrl']
```
zeantsoi: Do we want/need to set defaults for `adserver_click_url` and `adserver_impression_url` on `PromoCampaign`?

dwick: we don't need to.  i'm not super clear on when we add defaults, this isn't tied to the plugin so i suppose we could... @bsimpson63 thoughts?

bsimpson63: I think we've been not adding defaults for attributes that only exist in a plugin.

bsimpson63:
:fish:

dwick:
218c48ef405c57e5cc43c3ac342585144854a7ef 95c13f43263c5b2d21f7223685a3e42c02e3d214 79bc7cb34e32e7a595953e187d303e6e9253d2ee

PR #92, opened 2015-09-04
Support sponsored headlines coming directly from adzerk

dwick:
This will come after clicks goes out.

TODOS:
- [ ] sort out how clicks will be tracked without campaigns
- [ ] don't use reddit's traffic reporting for these


dwick:
likely be replacing this with a webhook.

PR #93, opened 2015-09-23
adzerk_api: Include response text in exception.

bsimpson63:
This will make it easier to figure out what's wrong just by looking at the adzerk_q logs.

:eyeglasses: @spladug @dwick


spladug:
:fish:

PR #94, opened 2015-09-25
Force bytestring when logging changes

zeantsoi:
:eyeglasses: @bsimpson63 @spladug 

Old problem recurring as documented in [JIRA ADS-260](https://reddit.atlassian.net/browse/ADS-260).

This change sanitizes all strings to bytestrings prior to sending to the logger. The operation only takes effect on strings containing non-ASCII characters; it is idempotent on strings comprised entirely on ASCII chars.

### DIFF
--- a/reddit_adzerk/adzerkpromote.py
+++ b/reddit_adzerk/adzerkpromote.py
@@ -108,7 +108,7 @@ def make_change_strings(changed):
             newval = date_from_adzerk(newval)
             oldval = date_from_adzerk(oldval)
 
-        return '%s: %s -> %s' % (attr, oldval, newval)
+        return '%s: %s -> %s' % (attr, _force_utf8(oldval), _force_utf8(newval))
 
     return map(change_to_str, changed)
 


reddit_adzerk/adzerkpromote.py
```diff
@@ -108,7 +108,7 @@ def change_to_str(change_tuple):
             newval = date_from_adzerk(newval)
             oldval = date_from_adzerk(oldval)
 
-        return '%s: %s -> %s' % (attr, oldval, newval)
+        return '%s: %s -> %s' % (attr, _force_utf8(oldval), _force_utf8(newval))
```
spladug: the alternate way of fixing this would be to leave out these force_utf8 calls and do `u'%s: %s -> %s'` then update the `log_text` stuff later on to also be unicode strings.  this is probably easier though and looks fine.

spladug:
:fish: from me, but probs want to wait for @bsimpson63 as well since he knows this code way better

bsimpson63:
:fish:

PR #96, opened 2015-09-29
Repair adzerk objects with incorrect ids

dwick:
Sometimes (we don't know why) adzerk will respond with an id that isn't a part
of our network.  In this case we should look up the object by name/relation and
set the id properly.  If we can't find it at all just unset it so a new object is
created.


reddit_adzerk/adzerk_api.py
```diff
@@ -7,16 +8,34 @@
...
+        if not isinstance(json_body, dict):
+            raise ValueError("invalid json")
+
+        if response.status_code == 400:
+            if RE_BAD_ID.match(json_body.get("message")):
+                raise AdzerkIdError()
     except ValueError:
-        raise AdzerkError('bad response')
+        if response.text and RE_BAD_ID.match(response.text):
+            raise AdzerkIdError()
```
bsimpson63: adzerk will give us a non-json response sometimes?

dwick: yes, they're returning 200s with `"... not in your network"`

json.loads("\"something\"") doesn't throw a value error, but it also just returns a unicode string.

reddit_adzerk/adzerkpromote.py
```diff
@@ -113,6 +115,84 @@ def change_to_str(change_tuple):
     return map(change_to_str, changed)
 
 
+def _repair_bad_creative_id(link, author):
+    fixed = False
+    if not getattr(author, "external_advertiser_id", False):
+        g.log.error("no author for link %s, cannot repair" % link._fullname)
+    else:
+        all_creatives = adzerk_api.Creative.list(author.external_advertiser_id)
```
bsimpson63: can you add a comment here (or somewhere else appropriate) explaining that the object exists in adzerk, but when we created it they told us the wrong id so we need to look it up.

reddit_adzerk/adzerkpromote.py
```diff
@@ -113,6 +115,84 @@ def change_to_str(change_tuple):
...
+    if not fixed:
+        g.log.info("unsetting creative id of %d for link %s" %
+            (link.external_creative_id, link._fullname))
+        link.external_creative_id = None
+        link._commit()
+
+
+def _repair_bad_campaign_id(link):
+    fixed = False
+    campaigns_by_fullname = adzerk_api.Campaign.search(link._fullname)
```
bsimpson63: `campaigns_by_fullname` sounds like it would be a dict

reddit_adzerk/adzerkpromote.py
```diff
@@ -113,6 +115,84 @@ def change_to_str(change_tuple):
...
+    for campaign in campaigns_by_fullname:
+        if campaign.Name == link._fullname:
+            if link.external_campaign_id != campaign.Id:
```
bsimpson63: why do you need this check? and why is it fixed if this is false and we don't set anything?

reddit_adzerk/adzerkpromote.py
```diff
@@ -113,6 +115,84 @@ def change_to_str(change_tuple):
...
+    if not fixed:
+        g.log.info("unsetting flight id of %d for campaign %s" %
+            (campaign.external_flight_id, campaign._fullname))
+        campaign.external_flight_id = None
+        campaign._commit()
+
+
+def _repair_bad_advertiser_id(author):
+    fixed = False
+    advertisers_by_name = adzerk_api.Advertiser.search(author.name)
```
bsimpson63: `advertisers_by_name` sounds like a dict also

bsimpson63:
:nail_care: looks good, just a couple comments

reddit_adzerk/adzerkpromote.py
dwick: sometimes we get `not in your network` but the id is actually correct, can't seem to tell when though

reddit_adzerk/adzerk_api.py
```diff
@@ -7,16 +8,34 @@
...
-        return json.loads(response.text)
+        json_body = json.loads(response.text)
+
+        if not isinstance(json_body, dict):
+            raise ValueError("invalid json")
+
+        if response.status_code == 400:
+            if RE_BAD_ID.match(json_body.get("message")):
+                raise AdzerkIdError()
     except ValueError:
```
zeantsoi: Is there an off-chance that `json.loads` will raise a `TypeError` if the `response.text` is empty?

dwick: doesn't look like it http://docs.python-requests.org/en/latest/api/#requests.Response.text

dwick:
:haircut:

bsimpson63:
:fish:

dwick:
closing since we hopefully no longer need this.

PR #97, opened 2015-10-01
Force adzerk to load specific ads based on a query param.

dwick:
:eyeglasses: @zeantsoi 

previously adzerk was hosting a lot of this code.  i moved it to the plugin to make changes like this easier (ie not involve adzerk).

### DIFF
--- a/reddit_adzerk/adzerkads.py
+++ b/reddit_adzerk/adzerkads.py
@@ -1,11 +1,19 @@
+import random
 from urllib import quote
 
 from pylons import tmpl_context as c
 from pylons import app_globals as g
+from pylons import request
 
+from r2.controllers import add_controller
+from r2.controllers.reddit_base import (
+    MinimalController,
+    UnloggedUser,
+)
+from r2.lib import promote
 from r2.lib.pages import Ads as BaseAds
+from r2.lib.wrapped import Templated
 from r2.models.subreddit import DefaultSR
-
 
 FRONTPAGE_NAME = "-reddit.com"
 
@@ -13,16 +21,60 @@ class Ads(BaseAds):
     def __init__(self):
         BaseAds.__init__(self)
 
-        url_key = "adzerk_https_url" if c.secure else "adzerk_url"
         site_name = getattr(c.site, "analytics_name", c.site.name)
 
         # adzerk reporting is easier when not using a space in the tag
         if isinstance(c.site, DefaultSR):
             site_name = FRONTPAGE_NAME
 
-        self.ad_url = g.config[url_key].format(
-            subreddit=quote(site_name.lower()),
+        placements = request.GET.get('placements', None)
+        sr = quote(site_name.lower())
+
+        ad_url = g.adzerk_url.format(
+            subreddit=sr,
             origin=c.request_origin,
             loggedin="loggedin" if c.user_is_loggedin else "loggedout",
         )
+
+        if placements:
+            ad_url = promote.update_query(ad_url, { "placements": placements })
+
+        self.ad_url = ad_url
         self.frame_id = "ad_main"
+
+
+class BaseAdFrame(Templated):
+    def __init__(self, **kwargs):
+        self.cache_buster = str(random.randint(0, 1000000))
+        return super(BaseAdFrame, self).__init__(**kwargs)
+
+
+class Ad300x250(BaseAdFrame):
+    pass
+
+
+class Ad300x250Companion(BaseAdFrame):
+    pass
+
+
+@add_controller
+class AdServingController(MinimalController):
+    def pre(self):
+        super(AdServingController, self).pre()
+
+        if request.host != g.media_domain:
+            # don't serve up untrusted content except on our
+            # specifically untrusted domain
+            self.abort404()
+
+        c.user = UnloggedUser([c.lang])
+        c.user_is_loggedin = False
+        c.forced_loggedout = True
+        c.allow_framing = True
+
+    def GET_ad_300_250(self):
+        return Ad300x250().render()
+
+    def GET_ad_300_250_companion(self):
+        return Ad300x250Companion().render()
+
--- a/reddit_adzerk/public/static/js/adzerk/adzerk.js
+++ b/reddit_adzerk/public/static/js/adzerk/adzerk.js
@@ -1,29 +1,39 @@
-r.adzerk = {
-    origin: location.protocol == 'https:'
-            ? 'https://static.adzerk.net'
-            : 'http://static.adzerk.net',
+!(function(window, $, undefined) {
+  'use strict';
 
-    createSponsorshipAdFrame: function() {
-        var iframe = $('<iframe>')
-            .attr({
-                'id': 'ad_sponsorship',
-                'src': r.adzerk.origin + '/reddit/ads-load.html?bust2',
-                'frameBorder': 0,
-                'scrolling': 'no'
-            })
-        $('.side .sponsorshipbox')
-            .empty()
-            .append(iframe)
-    }
-}
+  window.r = window.r || {};
 
-$(window).on('message', function(ev) {
-    ev = ev.originalEvent
-    if (ev.origin != r.adzerk.origin) {
-      return
+  r.adzerk = {
+
+    createSponsorshipAdFrame: function(overrideSrc) {
+      var $iframe = $('<iframe>');
+
+      $iframe
+        .attr({
+            id: 'ad_sponsorship',
+            src: '//' + r.config.media_domain + '/ads/display/300x250-companion',
+            frameBorder: 0,
+            scrolling: 'no',
+        });
+
+      $('.side .sponsorshipbox')
+        .empty()
+        .append($iframe);
+    },
+
+  };
+
+  $(window).on('message', function(e) {
+    e = e.originalEvent;
+
+    if (!new RegExp('^http(s)?:\\/\\/' + r.config.media_domain, 'i').test(e.origin)) {
+      return;
     }
-    msg = ev.data.split(':')
-    if (msg[0] == 'ados.createAdFrame') {
-      r.adzerk.createSponsorshipAdFrame()
+
+    var messsage = e.data.split(':')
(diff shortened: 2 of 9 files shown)


reddit_adzerk/public/static/js/adzerk/companion.js
```diff
@@ -0,0 +1,12 @@
+(function(undefined) {
+  'use strict';
+  var PLACEMENT = 'sponsorship';
+
+  window.onload = function() {
+    var adContent = window.parent.frames.ad_main.ados_results;
+    window.name = 'ad-' + PLACEMENT;
+    eval(adContent[PLACEMENT]);
```
JordanMilne: What is this `eval` meant to do? Is it necessary?

reddit_adzerk/templates/adbaseframe.html
```diff
@@ -0,0 +1,65 @@
...
+      color: #369;
+      display: block;
+      font-family: verdana, arial, sans-serif;
+      font-size: small;
+      text-decoration: none;
+      text-align: center;
+    }
+  </style>
+  <script>
+    window.ADS_GLOBALS = ${unsafe(json.dumps(config.get_js_config()))};
```
JordanMilne: This can be `scriptsafe_dumps()`, it won't break the wonky escaping that `get_js_config()` already does.

reddit_adzerk/public/static/js/adzerk/companion.js
dwick: unfortunately this is just how adzerk works.  i've moved the frames to the media domain/forced users to be logged out so i think that should cover our bases?

JordanMilne: Hmm, maybe make a comment that this is required because Adzerk sends JS payloads in their JSON. `eval()`s are pretty spooky.

dwick: yeah, i'd really not be doing this, but it's no change from how it currently works (just who serves it).  i'll speak to them to see if there are other options.  should be safe as long as it's sandboxed in the iframe though?

JordanMilne: Maybe, but I'd still make a note that explains why it's necessary, and that it won't run on the main domain.

dwick:
:eyeglasses: @madbook since it's basically all js (sorry!)

reddit_adzerk/adzerkads.py
```diff
@@ -16,13 +24,59 @@ def __init__(self):
         url_key = "adzerk_https_url" if c.secure else "adzerk_url"
         site_name = getattr(c.site, "analytics_name", c.site.name)
 
+
```
zeantsoi: Nit: double blanks should only be used to delimit class definitions.

reddit_adzerk/public/static/js/adzerk/adzerk.js
```diff
@@ -1,29 +1,39 @@
...
+
+    createSponsorshipAdFrame: function(overrideSrc) {
+      var $iframe = $('<iframe>');
+
+      $iframe
+        .attr({
+            'id': 'ad_sponsorship',
+            'src': '//' + r.config.media_domain + '/ads/display/300x250-companion',
+            'frameBorder': 0,
+            'scrolling': 'no'
```
madbook: nit: add a trailing comma here for diff-friendliness

reddit_adzerk/adzerkads.py
```diff
@@ -16,13 +24,59 @@ def __init__(self):
...
+            ad_url = promote.update_query(ad_url, { "placements": placements })
+
+        self.ad_url = ad_url
         self.frame_id = "ad_main"
+
+
+class BaseAdFrame(Templated):
+    def __init__(self, *args, **kwargs):
+        self.cache_buster = str(random.randint(0, 1000000))
+        return super(BaseAdFrame, self).__init__(*args, **kwargs)
```
madbook: Should remove `*args` from this.  As it is right now, I think this would actually throw an error if you tried to pass any non-keyword arguments in, since `Templated.__init__` doesn't support it.

reddit_adzerk/templates/adbaseframe.html
```diff
@@ -0,0 +1,67 @@
...
+          var id = window.parent.frames[frame].ados_ads[placement].id;
+          var discussLink = document.createElement('a');
+          var adzerkPreview = 'https://preview.adzerk.net/preview/' + id;
+
+          discussLink.className = 'discuss-ad';
+          discussLink.target = 'top';
+          discussLink.href = '//reddit.com/r/ads/submit?url=' + encodeURIComponent(adzerkPreview);
+          discussLink.innerHTML = 'discuss this ad on reddit';
+
+          document.getElementsByTagName('body')[0].appendChild(discussLink);
```
madbook: should be able to just do `document.body.appendChild(discussLink)` i think, unless that does something weird in frames that i can't remember

reddit_adzerk/adzerkads.py
```diff
@@ -16,13 +24,59 @@ def __init__(self):
...
+
+
+class Ad300x250Companion(BaseAdFrame):
+    pass
+
+
+@add_controller
+class AdServingController(MinimalController):
+    def pre(self, *args, **kwargs):
+        super(AdServingController, self).pre(*args, **kwargs)
```
madbook: similar to my other comment, i think `*args` and `**kwargs` should be removed here.  `MinimalController.pre` doesn't define them, so actually passing any extra arguments/keyword arguments here would throw an error. Passing them through here implies that they are supported.

madbook:
this is pretty much what's happening now, just moved so we're hosting it? i think the js stuff looks alright. it's a little tough to wrap my head around all this frame business :P  The only thing that stood out as a big concern for me was the `eval` use, but I see @JordanMilne's already brought that up.  Is this up on your staging?

:nail_care: a couple of minor comments/questions

reddit_adzerk/adzerkads.py
```diff
@@ -16,13 +24,59 @@ def __init__(self):
...
+class Ad300x250(BaseAdFrame):
+    pass
+
+
+class Ad300x250Companion(BaseAdFrame):
+    pass
+
+
+@add_controller
+class AdServingController(MinimalController):
```
zeantsoi: I'm guessing we don't want cached results... should we be overriding `try_pagecache`?

dwick:
yeah, the only real change is this: https://github.com/reddit/reddit-plugin-adzerk/pull/97/files#diff-a630e5a94de757d8008045ccc1f74b2eR35

everything else is a copy/cleanup of what adzerk was hosting. (i suppose i could make it 2 commits for clarity).  the frames are ridiculous:

1. the 300x250 ad requests both placements (300x250 and 300x100 companion) from with it's frame.
2. it renders the 300x250

if a companion comes back in the response as well:
1. it post messages up to the parent that we have one
2. the parent creates the frame for the 300x100
3. that frame contains js that looks up the 300x250 frame's `ados_results` and evals it to get the 300x100 to render

whomp whomp.

I think the ideal way might be to have the parent (reddit.com) hit the api (or hit our api that hits the api) and creates all the frames on the page instead of this weird sibling relationship that's going on right now. i'll likely get to that when we start framing sponsored headlines.

reddit_adzerk/adzerkads.py
dwick: not sure what that does, and it shouldnt matter that anything is cached since it's all requested via js from the client.  the only thing that might matter is the cache busting string on `ados.js`, which is what `self.cache_buster` is.  there's probably a better way to do that though.

reddit_adzerk/adzerkads.py
dwick: are you sure? i would think this would be the equivalent of `ParentClass.prototype.apply(this, arguments)`.  what if `Templated` changes, then we'd have to go update every subclass?  seems like that would be bad language design :-1: 

```
In [1]: class Foo(object):
    def __init__(self):                 
        print "foo'd"
   ...:         

In [2]: class Bar(Foo):
    def __init__(self, *args, **kwargs):
        super(Bar, self).__init__(*args, **kwargs)
        print "bar'd"
   ...:         

In [3]: Bar()
foo'd
bar'd
Out[3]: <__main__.Bar at 0x10442c450>

In [4]: Foo()
foo'd
Out[4]: <__main__.Foo at 0x10442c550>
```

dwick:
:haircut: general explanation above, not 100% sure about the `*args` `**kwargs` stuff though.  I put this up on https://dwick.staging.wireddit.com/

reddit_adzerk/adzerkads.py
zeantsoi: I think https://github.com/reddit/reddit-plugin-adzerk/pull/97#discussion_r41197884 makes a salient point: passing `*args` and `**kwargs` implicitly suggests that they are acceptable arguments when they _will_ throw an exception if passed. In the current implementation, no exception will be thrown, but it's not difficult to envision a scenario wherein an arg might be inadvertently passed in the future (e.g., I had no clue that `Templated` was implemented in the manner).

reddit_adzerk/adzerkads.py
zeantsoi: Remove extra line.

reddit_adzerk/adzerkads.py
madbook: The code snippet you posted is correct, the errors will only get thrown if you actually pass anything in, e.g. `Bar('some arg', some_kwargs=True)` would barf, though you wouldn't expect it to from looking at the `__init__` function

dwick: oh python, i remember having trouble with this: https://github.com/reddit/reddit-public/blob/df0b3b91580928b779757375c62ecd79c2ee3aa3/r2/r2/lib/validator/validator.py#L156.  now i know why..

dwick:
:haircut:

madbook:
:fish:

zeantsoi:
:fish:

PR #98, opened 2015-10-05
Remove static click trackers.

dwick:
These are no longer used.  Also they incorrectly were cached
permanently with no consideration for url updates.


dwick:
:eyeglasses: @zeantsoi

dwick:
see https://github.com/reddit/reddit-public/pull/4176/files for ref

zeantsoi:
:fish:

PR #99, opened 2015-10-05
adzerk_api.py: Fix exception on bad encoding

dwick:
:eyeglasses: @zeantsoi 

fixed in the package https://github.com/kennethreitz/requests/commit/2bb49ff386deab4e67e199450085720c1521411a (we're still on 0.12.1)


zeantsoi:
:fish:

PR #100, opened 2015-10-08
Cache harder

dwick:
:eyeglasses: @JordanMilne @spladug 

semi unrelated commit restricting the placements feature to sponsors, but i figured might as well deploy them both at once.

### DIFF
--- a/reddit_adzerk/adzerkads.py
+++ b/reddit_adzerk/adzerkads.py
@@ -1,11 +1,20 @@
+import json
+import random
 from urllib import quote
 
 from pylons import tmpl_context as c
 from pylons import app_globals as g
+from pylons import request
 
+from r2.controllers import add_controller
+from r2.controllers.reddit_base import (
+    MinimalController,
+    UnloggedUser,
+)
+from r2.lib import promote
 from r2.lib.pages import Ads as BaseAds
+from r2.lib.wrapped import Templated
 from r2.models.subreddit import DefaultSR
-
 
 FRONTPAGE_NAME = "-reddit.com"
 
@@ -19,9 +28,54 @@ class Ads(BaseAds):
         if isinstance(c.site, DefaultSR):
             site_name = FRONTPAGE_NAME
 
-        self.ad_url = g.adzerk_url.format(
-            subreddit=quote(site_name.lower()),
-            origin=c.request_origin,
-            loggedin="loggedin" if c.user_is_loggedin else "loggedout",
-        )
+        sr = quote(site_name.lower())
+        data = {
+            "keywords": ",".join([
+                sr,
+                "loggedin" if c.user_is_loggedin else "loggedout",
+            ]),
+            "origin": c.request_origin,
+        }
+
+        placements = request.GET.get("placements", None)
+
+        if placements:
+            data["placements"] = placements
+
+        self.ad_url = g.adzerk_url.format(data=json.dumps(data))
         self.frame_id = "ad_main"
+
+
+class BaseAdFrame(Templated):
+    pass
+
+
+class Ad300x250(BaseAdFrame):
+    pass
+
+
+class Ad300x250Companion(BaseAdFrame):
+    pass
+
+
+@add_controller
+class AdServingController(MinimalController):
+    def pre(self):
+        super(AdServingController, self).pre()
+
+        if request.host != g.media_domain:
+            # don't serve up untrusted content except on our
+            # specifically untrusted domain
+            self.abort404()
+
+        c.user = UnloggedUser([c.lang])
+        c.user_is_loggedin = False
+        c.forced_loggedout = True
+        c.allow_framing = True
+
+    def GET_ad_300_250(self):
+        return Ad300x250().render()
+
+    def GET_ad_300_250_companion(self):
+        return Ad300x250Companion().render()
+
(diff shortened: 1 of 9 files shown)


dwick:
example urls:

`https://dwick.staging-media.wireddit.com/ads/display/300x250#{"keywords": "movies,loggedin", "origin": "https://dwick.staging.wireddit.com"}`
`https://dwick.staging-media.wireddit.com/ads/display/300x250-companion`

any non-static data is contained in the hash as a query param, so CF should have no problems caching these.

the media domain server also needs to allow /ads/* to be served from the app servers.

reddit_adzerk/public/static/js/adzerk/companion.js
```diff
@@ -0,0 +1,19 @@
+(function(undefined) {
```
spladug: should this have `!`?

spladug: also is it correct to be defining `undefined` as `window`?

reddit_adzerk/public/static/js/adzerk/companion.js
```diff
@@ -0,0 +1,19 @@
...
+  'use strict';
+  var PLACEMENT = 'sponsorship';
+
+  window.onload = function() {
```
spladug: i think window should be coming from the function params, right?

reddit_adzerk/adzerkads.py
```diff
@@ -19,9 +28,54 @@ def __init__(self):
...
+            ]),
+            "origin": c.request_origin,
+        }
+
+        placements = request.GET.get("placements", None)
+
+        if c.user_is_sponsor and placements:
+            data["placements"] = placements
+
+        self.ad_url = g.adzerk_url.format(data=json.dumps(data))
```
spladug: is the json always going to be safe for URLs or do we need to percent encode it?

reddit_adzerk/public/static/js/adzerk/adzerk.js
```diff
@@ -1,29 +1,39 @@
...
+    e = e.originalEvent;
+
+    if (!new RegExp('^http(s)?:\\/\\/' + r.config.media_domain, 'i').test(e.origin)) {
+      return;
     }
-    msg = ev.data.split(':')
-    if (msg[0] == 'ados.createAdFrame') {
-      r.adzerk.createSponsorshipAdFrame()
+
+    var messsage = e.data.split(':')
```
spladug: :snake: messsssssssage

reddit_adzerk/adzerkads.py
```diff
@@ -19,9 +28,54 @@ def __init__(self):
...
         self.frame_id = "ad_main"
+
+
+class BaseAdFrame(Templated):
```
spladug: i think i'm missing something, why is this `BaseAdFrame` but the template is `adbaseframe.html`?

reddit_adzerk/adzerkads.py
```diff
@@ -19,9 +28,54 @@ def __init__(self):
         if isinstance(c.site, DefaultSR):
             site_name = FRONTPAGE_NAME
 
-        self.ad_url = g.adzerk_url.format(
-            subreddit=quote(site_name.lower()),
-            origin=c.request_origin,
-            loggedin="loggedin" if c.user_is_loggedin else "loggedout",
-        )
+        sr = quote(site_name.lower())
```
spladug: just to confirm, for something like a multi this won't be multiple subreddit names right? if it is multiple names should they be sorted for cache key defragmentation?

reddit_adzerk/adzerkads.py
spladug: we might want to do `sort_keys=True` to ensure the cache key is maximally similar; as-is it could vary at the whim of the python dict right?

spladug:
jordan's out today (canadian thanksgiving) and i think the presence of `eval` and untrusted code makes me wary enough that i'd like to ask that we hold this til he's had time to review, sorry.

spladug:
:nail_care: mostly minor stuff. just some cache key concerns and styleguide nits.  thanks for this re-do.

dwick:
previous discussion on `eval`: https://github.com/reddit/reddit-plugin-adzerk/pull/97#discussion-diff-40982786

reddit_adzerk/adzerkads.py
dwick: whoops, if these are named properly do i not need the `<%inherit file="adbaseframe.html"/>`?

reddit_adzerk/adzerkads.py
dwick: these params go in the fragment: https://github.com/reddit/reddit-private/pull/1059/files. never sent to the server so order shouldnt be a problem.

reddit_adzerk/adzerkads.py
spladug: no, you'll still need the mako-level inheritance unfortunately.

reddit_adzerk/adzerkads.py
spladug: ah right, cool

dwick:
:haircut:

spladug:
:fish:

reddit_adzerk/adzerkads.py
JordanMilne: If the JSON never has percents in it we should be fine, otherwise we'll need to use weird hacks like https://github.com/reddit/reddit-public/pull/4187/files#diff-123b9fad9beeb0330a01dab96176f0a4R31

JordanMilne:
:fish:

PR #101, opened 2015-10-15
Add support for native

dwick:
:eyeglasses: @zeantsoi 

https://reddit.atlassian.net/browse/ADS-271

### DIFF
--- a/reddit_adzerk/adzerkpromote.py
+++ b/reddit_adzerk/adzerkpromote.py
@@ -34,6 +34,7 @@ from r2.lib.validator import (
     validate,
     VPrintable,
     VBoolean,
+    VOneOf,
 )
 
 from r2.models import (
@@ -258,6 +259,11 @@ def update_flight(link, campaign, az_campaign):
     else:
         az_flight = None
 
+    # backwards compatability
+    if campaign.platform == "mobile":
+        campaign.platform = "mobile_web"
+        campaign._commit()
+
     campaign_overdelivered = is_overdelivered(campaign)
     delayed_start = campaign.start_date + datetime.timedelta(minutes=15)
     if delayed_start >= campaign.end_date:
@@ -344,22 +350,12 @@ def update_flight(link, campaign, az_campaign):
         })
 
     if campaign.platform != 'all':
-        siteZones = []
-        if campaign.platform == 'desktop':
-            siteZones.append({
-                'SiteId': g.az_selfserve_site_id,
+        d.update({
+            'SiteZoneTargeting': [{
+                'SiteId': g.az_selfserve_site_ids[campaign.platform],
                 'IsExclude': False,
-            })
-        elif campaign.platform == 'mobile':
-            siteZones.append({
-                'SiteId': g.az_selfserve_mobile_web_site_id,
-                'IsExclude': False,
-            })
-
-        if len(siteZones):
-            d.update({
-                'SiteZoneTargeting': siteZones
-            })
+            }],
+        })
 
     # special handling for location conversions between reddit and adzerk
     if campaign.location:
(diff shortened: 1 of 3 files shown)


reddit_adzerk/adzerkpromote.py
```diff
@@ -711,21 +702,30 @@ class AdzerkApiController(api.ApiController):
...
+        platform=VOneOf("platform", [
+            "desktop",
+            "mobile_web",
+            "mobile_native",
+        ], default=None),
         loid=nop('loid', None),
         is_refresh=VBoolean("is_refresh", default=False),
     )
-    def POST_request_promo(self, srnames, is_mobile_web, loid, is_refresh):
+    def POST_request_promo(self, srnames, is_mobile_web, site, loid, is_refresh):
```
dwick: `site` should be `platform`

reddit_adzerk/adzerkpromote.py
```diff
@@ -711,21 +702,30 @@ class AdzerkApiController(api.ApiController):
     @validate(
         srnames=VPrintable("srnames", max_length=2100),
         is_mobile_web=VBoolean('is_mobile_web'),
+        platform=VOneOf("platform", [
```
zeantsoi: Rather than dealing with the semantics, is it better to define this as an `Enum`?

reddit_adzerk/adzerkpromote.py
zeantsoi: Nit: I worry that we may conflate `site` with `platform`, as we use the two interchangeably in this instance, but distinctly in other situations.

reddit_adzerk/adzerkpromote.py
```diff
@@ -238,6 +239,11 @@ def update_flight(link, campaign, az_campaign):
     else:
         az_flight = None
 
+    # backwards compatability
```
zeantsoi: Will we run a migration for the _live_ campaigns that are `mobile`? Otherwise, I think they might be in a purgatory where there's no `SiteId` corresponding to the platform for the remainder of their lifetime.

zeantsoi:
:nail_care:

reddit_adzerk/adzerkpromote.py
dwick: no, `mobile` is aliased to `mobile_web`, we can remove this eventually

dwick: to be clear, the actual site id won't be changing for mobile web

reddit_adzerk/adzerkpromote.py
dwick: renamed to site

dwick: although we currently map platforms to site ids, not sure there is a difference?

dwick:
:haircut:

reddit_adzerk/adzerkpromote.py
zeantsoi: I'm still a little worried about the nomenclature here, although I see what you're getting at (i.e., site ids mapping to platform names).

The primary issue, IMO, is that the term _site_ is variously defined to mean one of at least three very distinct things:

1. current page
2. device/platform
3. SiteId for Adzerk

At first blush, the term _site_ doesn't intuitively invoke `SiteId`, which is how it's used in this case. Maybe there's a better way of naming this entirely? If not, I think it'd be helpful to comment or name the arg more clearly.

zeantsoi:
:nail_care: 

One nit: tentative fish.

reddit_adzerk/adzerkpromote.py
dwick: i think it should be platform since that's what we use elsewhere, including on the model.

dwick:
:haircut:

zeantsoi:
:fish:

dwick:
:haircut: rebased and found some issues with the config

zeantsoi:
Few comments, but :fish:

PR #102, opened 2015-10-15
Display ads: Send the same keywords as sponsored headlines

dwick:
:eyeglasses: @zeantsoi


reddit_adzerk/adzerkads.py
```diff
@@ -22,16 +24,14 @@ class Ads(BaseAds):
     def __init__(self):
         BaseAds.__init__(self)
 
-        site_name = getattr(c.site, "analytics_name", c.site.name)
+        srnames = promote.srnames_from_site(c.user, c.site)
 
         # adzerk reporting is easier when not using a space in the tag
         if isinstance(c.site, DefaultSR):
-            site_name = FRONTPAGE_NAME
+            srnames = srnames - {Frontpage.name} | {FRONTPAGE_NAME}
```
zeantsoi: If I'm not mistaken, Adzerk maintains a distinction between ` reddit.com` (with a space) and `-reddit.com`. One denotes display ads and the other native ads.

Does this change have any implication on that?

dwick: yeah, i'm not sure why they're different for display/native, but this is removing ` reddit.com` if it's there and adding in `-reddit.com` instead. we should probably figure out if we can make these the same, it might be something legacy.

```
>>> {" reddit.com"} - {" reddit.com"} | {"-reddit.com"}
set(['-reddit.com'])
```

zeantsoi: Yah, [here's the discussion](https://adzerk.slack.com/archives/adzerk-reddit/p1444782391000118). Sounds like we'll need to coordinate with Adzerk to ensure the mapping remains correct.

dwick: there's a ticket to change this all together: https://reddit.atlassian.net/browse/ADS-279

for now this is no change.

dwick:
:haircut:

zeantsoi:
:fish:

dwick:
bf115a14b1e5f695a3ac99b3b51b05d1aa63a6ee

PR #106, opened 2015-10-30
Fix ssl for preview links

dwick:
:eyeglasses: @zeantsoi 

From james:
> Can you change your client script to use preview.adzerk.com instead of preview.adzerk.net? I tried to add an SSL cert for preview.adzerk.net but they don't make a cert that will let us do preview.adzerk.net and *.adzerk.com (or any multiple wildcard domains).


zeantsoi:
:fish:

PR #108, opened 2015-11-09
Ads reporting: Add jobs to run/process link and campaign reports.

dwick:
- Campaign reports are lifetime totals.
 - Link reports get a daily breakdown by campaign.

Hourly job: reddit-job-queue_promo_reports.conf
Queue processor: reddit-consumer-adzerk_reporting_q.conf

:eyeglasses: @bsimpson63 @zeantsoi 

puppet stuff coming soon

### DIFF
--- a/reddit_adzerk/adzerkreporting.py
+++ b/reddit_adzerk/adzerkreporting.py
@@ -1,0 +1,450 @@
+"""
+Generate and processes daily/lifetime reports for ad campaigns.
+
+Runs link reports by day, by campaign, and stores in traffic db.
+Runs campaign reports for lifetime impression, clicks, and spend.
+
+A cron job is used to queue reports for promos that are currently
+serving, or served the day before.  The queue first makes all the
+requests to generate the reports before trying to retrieve any.
+This should give enough time for the first report to finish and
+be available by the time reading begins.
+"""
+
+import itertools
+import json
+import pytz
+import time
+from collections import defaultdict
+from datetime import datetime, timedelta
+
+from dateutil.parser import parse as parse_date
+from pylons import app_globals as g
+from sqlalchemy.orm import scoped_session, sessionmaker
+
+from r2.lib import (
+    amqp,
+    promote,
+)
+from r2.models import (
+    Link,
+    PromoCampaign,
+)
+from r2.models.traffic import (
+    engine,
+    AdserverClickthroughsByCodename,
+    AdserverImpressionsByCodename,
+    AdserverSpentByCodename,
+    AdserverTargetedClickthroughsByCodename,
+    AdserverTargetedImpressionsByCodename,
+    AdserverTargetedSpentByCodename,
+)
+
+from reddit_adzerk import (
+    adzerk_api,
+    report,
+)
+
+
+Session = scoped_session(sessionmaker(bind=engine))
+
+
+def queue_promo_reports():
+    """
+    Queue reports for promos that are currently
+    serving, or served the day before.
+    """
+    prev_promos = promote.get_served_promos(offset=-1)
+    promos = promote.get_served_promos(offset=0)
+    already_processed_links = set()
+    already_processed_campaigns = set()
+
+    for campaign, link in itertools.chain(prev_promos, promos):
+        if link._id36 not in already_processed_links:
+            _generate_link_report(link)
+            already_processed_links.add(link._id36)
+
+        if campaign._id36 not in already_processed_campaigns:
+            _generate_promo_report(campaign)
+            already_processed_campaigns.add(campaign._id36)
+
+
+def _generate_link_report(link):
+    amqp.add_item("adzerk_reporting_q", json.dumps({
+        "action": "generate_daily_link_report",
+        "link_id": link._id,
+    }))
+
+
+def _generate_promo_report(campaign):
+    amqp.add_item("adzerk_reporting_q", json.dumps({
+        "action": "generate_lifetime_campaign_report",
+        "campaign_id": campaign._id,
+    }))
+
+
+def _trigger_link_report(link, report_id, queued_date):
+    amqp.add_item("adzerk_reporting_q", json.dumps({
+        "action": "daily_link_report",
+        "link_id": link._id,
+        "report_id": report_id,
+        "queued_date": queued_date.isoformat(),
+    }))
+
+
+def _trigger_campaign_report(campaign, report_id, queued_date):
+    amqp.add_item("adzerk_reporting_q", json.dumps({
+        "action": "lifetime_campaign_report",
+        "campaign_id": campaign._id,
+        "report_id": report_id,
+        "queued_date": queued_date.isoformat(),
+    }))
+
+
+def _normalize_usage(impressions, clicks, spent):
+    # adzerk processes clicks faster than impressions
+    # throw away results that are obviously wrong.
+    if clicks > impressions:
+        impressions = 0
+        clicks = 0
+        spent = 0
+
+    return (impressions, clicks, spent)
+
+
+def _get_total_impressions(report_fragment):
+    return report_fragment.get("TotalImpressions", 0)
+
+
+def _get_total_clicks(report_fragment):
+    return (report_fragment.get("TotalClicks", 0) -
+        report_fragment.get("TotalFraudulentClicks", 0))
+
+
+def _get_total_spent(report_fragment):
+    return report_fragment.get("TotalTrueRevenue", 0)
+
+
+def _get_total_usage(report_fragment):
+    impressions = _get_total_impressions(report_fragment)
+    clicks = _get_total_clicks(report_fragment)
+    spent = _get_total_spent(report_fragment)
+
+    return _normalize_usage(impressions, clicks, spent)
+
+
+def _get_impressions(report_fragment):
+    return report_fragment.get("Impressions", 0)
+
+
+def _get_clicks(report_fragment):
+    return (report_fragment.get("Clicks", 0) -
+        report_fragment.get("FraudulentClicks", 0))
+
+
+def _get_spent(report_fragment):
+    return report_fragment.get("TrueRevenue", 0)
+
+
+def _get_usage(report_fragment):
+    impressions = _get_impressions(report_fragment)
+    clicks = _get_clicks(report_fragment)
+    spent = _get_spent(report_fragment)
+
+    return _normalize_usage(impressions, clicks, spent)
+
+def _get_date(report_fragment):
+    date = report_fragment.get("Date")
+
+    if not date:
+        return None
+
+    return parse_date(date)
+
+
+def _get_fullname(cls, report_fragment):
+    fullname = report_fragment.get("Title", "")
+
+    if not fullname.startswith(cls._fullname_prefix):
+        return None
+    else:
+        return fullname
+
+def _get_flight_id(report_fragment):
+    return report_fragment.get("Grouping", {}).get("OptionId", None)
+
+
+def _handle_generate_daily_link_report(link_id):
+    now = datetime.utcnow()
+    link = Link._byID(link_id, data=True)
+    campaigns = PromoCampaign._by_link(link._id)
+
+    link_start = min([promo.start_date for promo in campaigns])
+    link_end = max([promo.end_date for promo in campaigns])
+
+    now = now.replace(tzinfo=pytz.utc)
+    link_start = link_start.replace(tzinfo=pytz.utc)
+    link_end = link_end.replace(tzinfo=pytz.utc)
+
+    # if data has already been processed then there's no need
+    # to redo it.  use the last time the report was run as a 
+    # starting point, but subtract 24hrs since initial numbers
+    # are preliminary.
+    if getattr(link, "last_daily_report_run") is not None:
+        start = max([
+            link.last_daily_report_run - timedelta(hours=24),
+            link_start,
+        ])
+
+    else:
+        start = link_start
+
+    end = min([now, link_end])
+
+    report_id = report.queue_report(
+        start=start,
+        end=end,
+        groups=["optionId", "day"],
+        parameters=[{
+            "campaignId": link.external_campaign_id,
+        }],
+    )
+
+    _trigger_link_report(
+        link=link,
+        report_id=report_id,
+        queued_date=now,
+    )
+
+
+def _handle_generate_lifetime_campaign_report(campaign_id):
+    now = datetime.utcnow()
+    campaign = PromoCampaign._byID(campaign_id, data=True)
+    start = campaign.start_date.replace(tzinfo=pytz.utc)
+    end = campaign.end_date.replace(tzinfo=pytz.utc)
+    now = now.replace(tzinfo=pytz.utc)
+
+    end = min([now, end])
+
+    report_id = report.queue_report(
+        start=start,
+        end=end,
+        parameters=[{
+            "flightId": campaign.external_flight_id,
+        }],
+    )
+
+    _trigger_campaign_report(
+        campaign=campaign,
+        report_id=report_id,
+        queued_date=now,
+    )
+
+
+def _handle_lifetime_campaign_report(campaign_id, report_id, queued_date):
+    campaign = PromoCampaign._byID(campaign_id, data=True)
+
+    try:
+        report_result = report.fetch_report(report_id)
+    except report.ReportPendingException as e:
+        time.sleep(1)
+        raise
+    except report.ReportFailedException as e:
+        g.log.error(e)
+
+        # retry if report failed
+        _generate_promo_report(campaign)
+        return
+
+    impressions, clicks, spent = _get_total_usage(report_result)
+
+    campaign.adserver_spent_pennies = int(spent * 100)
+    campaign.adserver_impressions = impressions
+    campaign.adserver_clicks = clicks
+    campaign.last_lifetime_report = report_id
+    campaign.last_lifetime_report_run = queued_date
+
+    campaign._commit()
+
+
+def _handle_daily_link_report(link_id, report_id, queued_date):
+    link = Link._byID(link_id, data=True)
+
+    try:
+        report_result = report.fetch_report(report_id)
+    except report.ReportPendingException as e:
+        time.sleep(1)
+        raise
+    except report.ReportFailedException as e:
+        g.log.error(e)
+
+        # retry if report failed
+        _generate_link_report(link)
+        return
+
+    g.log.debug(report_result)
+
+    campaigns_by_fullname = {campaign._fullname: campaign for campaign
+        in PromoCampaign._by_link(link._id)}
+
+    # report is by date, by flight. each record is a day
+    # and each detail is a flight for that day.
+    for record in report_result.get("Records", []):
+        impressions, clicks, spent = _get_usage(record)
+        date = _get_date(record)
+
+        _insert_daily_link_reporting(
+            codename=link._fullname,
+            date=date,
+            impressions=impressions,
+            clicks=clicks,
+            spent=spent,
+        )
+
+        for detail in record.get("Details", []):
+            campaign_fullname = _get_fullname(PromoCampaign, detail)
+
+            if not campaign_fullname:
+                g.log.error("invalid fullname for campaign (%s/%s)" %
+                    (campaign_fullname, flight_id))
+                continue
+
+            campaign = campaigns_by_fullname.get(campaign_fullname)
+
+            if not campaign:
+                flight_id = _get_flight_id(detail)
+                g.log.warning("no campaign for flight (%s/%s)" %
+                    (campaign_fullname, flight_id))
+                continue
+
+            impressions, clicks, spent = _get_usage(detail)
+
+            _insert_daily_campaign_reporting(
+                codename=campaign._fullname,
+                date=date,
+                impressions=impressions,
+                clicks=clicks,
+                spent=spent,
+                subreddit=campaign.target_name,
+            )
+
+    link.last_daily_report = report_id
+    link.last_daily_report_run = queued_date
+    link._commit()
+
+
+def process_report_q():
+    @g.stats.amqp_processor('adzerk_q')
+    def _processor(message):
+        data = json.loads(message.body)
+        action = data.get("action")
+
+        if action == "daily_link_report":
+            _handle_daily_link_report(
+                link_id=data.get("link_id"),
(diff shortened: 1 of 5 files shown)


bsimpson63:
Why a queue rather than a cron?

bsimpson63:
Do you need to do any special handling to account for our weird way of handling dates for ads (campaigns start at midnight eastern--I forget the actual time/timezone used).

dwick:
the queue simplifies failures, retries, etc.  it should also be easier to scale by adding new consumers

dwick:
adzerk reporting is entirely in GMT, I have [this](https://github.com/reddit/reddit-plugin-adzerk/pull/108/files#diff-a369a8ee1a6bfac28a8b167eb9f58fa7R226), but not entirely sure it's necessary.

bsimpson63:
So will that have the expected/desired behavior? Say I have an ad running from 11/10-11/12. On reddit it will start at 00:00 EST on 11/10 and end at 00:00 EST on 11/13. If adzerk is reporting in GMT won't we have mismatched "day" buckets?

bsimpson63:
Have you tested this with adzerk? Is their reporting fast enough? What are the implications if adzerk slows down or is unreachable?

dwick:
I've tested locally, but it's kind of a pain to copy data from prod to my computer and fix all the ids.  Seems pretty fast with the amount of data we pull in any one report.  The plan would be to have the job/queue going for a bit before we make a decision on how often to run them and if we're satisfied enough with it to launch.

I've also confirmed with adzerk that there's no way to override the timezone for bucketing until they get their next version out (6-8 weeks).  I think maybe just adding a note that days are bucketed in UTC would be fine for now.  I'm not sure if anyone really cares about the daily/hourly breakdown aside from a debugging standpoint.

dwick:
After talking with people from sales we've decided to just add a message on the UI side of things about the GMT buckets.

reddit_adzerk/adzerkreporting.py
```diff
@@ -0,0 +1,345 @@
...
+
+def _ctr(clicks, impressions):
+    if not impressions:
+        return 0
+
+    return clicks / (impressions * 1.0) * 100
+
+
+def queue_promo_reports():
+    promos = promote.get_served_promos()
```
bsimpson63: Just checking, but if an ad is live but hasn't served any impressions we won't get impressions/clicks from adzerk. Will the traffic page will do the right thing?

reddit_adzerk/adzerkreporting.py
```diff
@@ -0,0 +1,345 @@
+import json
+import time
+from collections import defaultdict
+from datetime import datetime, timedelta
+from dateutil.parser import parse as parse_date
+from pylons import app_globals as g
```
bsimpson63: I think we usually have separate sections for built in python libraries and 3rd party ones.

reddit_adzerk/adzerkreporting.py
bsimpson63: Do you also need to get served promos from the previous day? Otherwise we'll miss the final update if adzerk is a little slow or if we check the report before the end of the hour.

reddit_adzerk/adzerkreporting.py
```diff
@@ -0,0 +1,345 @@
...
+def _ctr(clicks, impressions):
+    if not impressions:
+        return 0
+
+    return clicks / (impressions * 1.0) * 100
+
+
+def queue_promo_reports():
+    promos = promote.get_served_promos()
+    daily_queued = {}
```
bsimpson63: you cound use a `set` here

reddit_adzerk/adzerkreporting.py
```diff
@@ -0,0 +1,345 @@
...
+
+def _queue_promo_report(campaign):
+    amqp.add_item("adzerk_reporting_q", json.dumps({
+        "action": "queue_lifetime_campaign_report",
+        "campaign_id": campaign._id,
+    }))
+
+
+def process_report_q():
+    def _handle_lifetime_campaign_report(campaign_id, report_id, report_queued):
```
bsimpson63: why are these functions defined within `process_report_q`?

reddit_adzerk/adzerkreporting.py
```diff
@@ -0,0 +1,345 @@
...
+    def _handle_queue_daily_link_report(link_id):
+        now = datetime.utcnow()
+        link = Link._byID(link_id, data=True)
+        campaigns = PromoCampaign._by_link(link._id)
+
+        link_start = min([promo.start_date for promo in campaigns])
+        link_end = max([promo.end_date for promo in campaigns])
+
+        # make dates in the same timezone
+        now = now.replace(tzinfo=link_start.tzinfo)
```
bsimpson63: is this not always UTC?

reddit_adzerk/adzerkreporting.py
```diff
@@ -0,0 +1,345 @@
...
+
+
+def process_report_q():
+    def _handle_lifetime_campaign_report(campaign_id, report_id, report_queued):
+        report_data = report.fetch_report(report_id)
+        status = report_data["Status"]
+
+        if status == report.STATUS.PENDING:
+            time.sleep(1)
+            raise ValueError("report pending for campaign (%d/%s)" %
```
bsimpson63: This stuff seems to be not well abstracted/encapsulated. Perhaps `report.fetch_report` could raise an exception if the report was pending, and a different exception if there was an error and then you do what you need to do here based on that. Also, rather than relying on knowledge of the adzerk report format you could return a `namedtuple` with just the fields we care about.

reddit_adzerk/adzerkreporting.py
dwick: good call.

reddit_adzerk/adzerkreporting.py
dwick: reports are so we probably want to convert everything to UTC if it isn't. are campaign dates UTC or something else?

reddit_adzerk/adzerkreporting.py
dwick: > Just checking, but if an ad is live but hasn't served any impressions we won't get impressions/clicks from adzerk. Will the traffic page will do the right thing?

Everything is zeroed out like it is now.  example: https://www.reddit.com/traffic/3qroxn/

reddit_adzerk/adzerkreporting.py
dwick: is this a bad practice? just seemed like a decent way to separate the handlers from everything else

dwick:
:haircut: 
have a few questions

reddit_adzerk/adzerkreporting.py
bsimpson63: it's not really necessary, and the extra level of indentation makes it tougher to read

reddit_adzerk/adzerkreporting.py
bsimpson63: how about just force it to UTC rather than whatever the `link_start` tz happens to be

reddit_adzerk/adzerkreporting.py
```diff
@@ -0,0 +1,336 @@
...
+
+        try:
+            report_result = report.fetch_report(report_id)
+        except report.ReportPendingException as e:
+            time.sleep(1)
+            raise
+        except report.ReportFailedException as e:
+            g.log.error(e)
+
+            # retry if report failed
```
bsimpson63: Any idea what could happen that would make a report fail? Is it something that makes sense to retry?

reddit_adzerk/adzerkreporting.py
```diff
@@ -0,0 +1,336 @@
...
+            start=start,
+            end=end,
+            groups=["optionId", "day"],
+            parameters=[{
+                "campaignId": link.external_campaign_id,
+            }],
+        )
+
+        amqp.add_item("adzerk_reporting_q", json.dumps({
+            "action": "daily_link_report",
```
bsimpson63: It took me a while to understand that there are separate "queue for report generation" and "read generated report" steps. The naming is additionally confusing because we're interacting with a reddit queue as well. Can you try to clean up these names? Also it would be good to put these "read generated report" steps into their own top level functions like the other places where you're calling `amqp.add_item`.

reddit_adzerk/adzerkreporting.py
```diff
@@ -0,0 +1,336 @@
+import json
```
bsimpson63: Can you add a module level docstring and/or function docstrings describing how this all works?

reddit_adzerk/adzerkreporting.py
```diff
@@ -0,0 +1,336 @@
...
+
+
+def process_report_q():
+    def _handle_lifetime_campaign_report(campaign_id, report_id, report_queued):
+        campaign = PromoCampaign._byID(campaign_id, data=True)
+
+        try:
+            report_result = report.fetch_report(report_id)
+        except report.ReportPendingException as e:
+            time.sleep(1)
```
bsimpson63: Without knowing how adzerk handles queuing/report generation, do you think there are any possibilities for weird race conditions or inefficiencies by having a single queue on our end to handle both requesting that a report be generated AND retrieving generated reports?

I guess in practice we'll be making all the requests to generate the reports, and then all the requests to retrieve the reports, which sounds like the right way to do it. Would it be safer to have separate queues for the two operations so that we're not blocking making requests for new reports while waiting to retrieve pending reports? Or does that add too much complexity?

reddit_adzerk/adzerkreporting.py
```diff
@@ -0,0 +1,336 @@
...
+            raise
+        except report.ReportFailedException as e:
+            g.log.error(e)
+
+            # retry if report failed
+            _queue_promo_report(campaign)
+            return
+
+        impressions = report_result.get("TotalImpressions", 0)
+        clicks = report_result.get("TotalClicks", 0) - report_result.get("TotalFraudulentClicks", 0)
```
bsimpson63: this line is too long!

reddit_adzerk/adzerkreporting.py
```diff
@@ -0,0 +1,336 @@
...
+
+        # adzerk processes clicks faster than impressions
+        if clicks > impressions:
+            return
+
+        campaign.adserver_spent = report_result.get("TotalTrueRevenue", 0)
+        campaign.adserver_impressions = impressions
+        campaign.adserver_clicks = clicks
+        campaign.last_lifetime_report = report_id
+        campaign.last_lifetime_report_run = report_queued
```
bsimpson63: is `report_queued` a date? if so can you use a variable name that indicates that?

reddit_adzerk/adzerkreporting.py
```diff
@@ -0,0 +1,336 @@
...
+        impressions = report_result.get("TotalImpressions", 0)
+        clicks = report_result.get("TotalClicks", 0) - report_result.get("TotalFraudulentClicks", 0)
+
+        # adzerk processes clicks faster than impressions
+        if clicks > impressions:
+            return
+
+        campaign.adserver_spent = report_result.get("TotalTrueRevenue", 0)
+        campaign.adserver_impressions = impressions
+        campaign.adserver_clicks = clicks
```
bsimpson63: Where is `adserver_clicks` used? Why doesn't this method write to the traffic db?

reddit_adzerk/adzerkreporting.py
```diff
@@ -0,0 +1,336 @@
...
+            # retry if report failed
+            _queue_link_report(link)
+            return
+
+        g.log.debug(report_data)
+
+        campaigns_by_fullname = {campaign._fullname: campaign for campaign
+            in PromoCampaign._by_link(link._id)}
+
+        session = scoped_session(sessionmaker(bind=engine))
```
bsimpson63: Elsewhere this is done at module level. I think that makes more sense organizationally since you're instantiating it here and then passing it into the method that actually writes to the db.

reddit_adzerk/adzerkreporting.py
```diff
@@ -0,0 +1,336 @@
...
+
+        for record in report_result.get("Records", []):
+            impressions = record.get("Impressions", 0)
+            clicks = record.get("Clicks", 0) - record.get("FraudulentClicks", 0)
```
bsimpson63: It would be nice to convert the adzerk report into the form we will want to use in `fetch_report`. That way you could: do conversions like this (excluding fraudulent clicks), include just the entries we care about, and provide sensible defaults for missing values.

reddit_adzerk/adzerkreporting.py
```diff
@@ -0,0 +1,336 @@
...
+
+            # adzerk processes clicks faster than impressions
+            if clicks > impressions:
+                continue
```
bsimpson63: Is this safe? Couldn't we end up only partially processing a report?

reddit_adzerk/adzerkreporting.py
```diff
@@ -0,0 +1,336 @@
...
+            start = link.last_daily_report_run - timedelta(hours=24)
+        else:
+            start = link_start
+
+        end = min([now, link_end])
+
+        report_id = report.queue_report(
+            start=start,
+            end=end,
+            groups=["optionId", "day"],
```
bsimpson63: Is there any way to make it so the caller doesn't need to know about adzerk specific settings like optionId and campaignId?

reddit_adzerk/adzerkreporting.py
```diff
@@ -0,0 +1,336 @@
...
+        clicks, ctr, spent):
+
+    date_formatted = date.strftime("%Y-%m-%d 00:00:00")
+    clicks_row = AdserverClickthroughsByCodename(**{
+        "codename": codename,
+        "date": date_formatted,
+        "interval": "day",
+        "unique_count": clicks,
+        "pageview_count": clicks,
+    })
```
bsimpson63: any reason to use `AdserverClickthroughsByCodename(**{})` rather than `AdserverClickthroughsByCodename(codename=codename, ...)`?

reddit_adzerk/adzerkreporting.py
```diff
@@ -0,0 +1,336 @@
...
+            g.log.warning("adzerk_reporting_q: unknown action - \"%s\"" % action)
+
+    amqp.consume_items("adzerk_reporting_q", _processor, verbose=False)
+
+
+def _insert_daily_link_reporting(
+        session, codename, date, impressions,
+        clicks, ctr, spent):
+
+    date_formatted = date.strftime("%Y-%m-%d 00:00:00")
```
bsimpson63: You should pass in a `datetime` object like done here https://github.com/reddit/reddit-public/blob/master/r2/r2/models/subreddit.py#L1482

bsimpson63:
:nail_care:

reddit_adzerk/adzerkreporting.py
dwick: i dont see much benefit from abstracting out something that will only ever be used with adzerk and requires knowledge of adzerk. it's also basically the same thing as [this](https://github.com/reddit/reddit-plugin-adzerk/blob/8e048acf4372565bf3d622f5ff5783965fb6156a/reddit_adzerk/report.py#L53) without the defaults and uses the async endpoint

reddit_adzerk/adzerkreporting.py
dwick: it should only be skipping the newest record (day).  the alternative is having clicks be a larger number than impressions, which imo is way worse.

reddit_adzerk/adzerkreporting.py
dwick: the format of the report is different based on the arguments used to queue it.

reddit_adzerk/adzerkreporting.py
dwick: easier access on the edit pages.

reddit_adzerk/adzerkreporting.py
dwick: it's rare and not an idempotent state according to james from adzerk.

reddit_adzerk/adzerkreporting.py
dwick: > in practice we'll be making all the requests to generate the reports, and then all the requests to retrieve the reports

this is what i was going for.  queuing all the reports should give enough time for the first report to finish and be available by the time we get to reading.  by doing it this way we're essentially having a single process polling on `ReportPendingException`s and not inadvertently slamming adzerk also.  if we add more queue processors i'd probably increase the sleep time here.

reddit_adzerk/adzerkreporting.py
dwick: was looking at this: https://github.com/reddit/reddit-public/blob/7227c7c468e00d01bf9b98b63470528a077e77ef/r2/r2/lib/traffic/traffic.py#L173

if we do it at the module level is there a connection open that's only used by the cron job?

reddit_adzerk/adzerkreporting.py
dwick: > Also it would be good to put these "read generated report" steps into their own top level functions like the other places where you're calling amqp.add_item.

I'm not sure that's entirely useful since it requires a queue action to trigger a read action

dwick:
:haircut:

reddit_adzerk/adzerkreporting.py
dwick: done

reddit_adzerk/adzerkreporting.py
bsimpson63: All these comments are coming from a readability standpoint. This module is really only doing one thing but with a lot of moving parts. Keeping things organized and in recognizable patterns helps when someone comes along fresh and tries to understand what is happening.

reddit_adzerk/adzerkreporting.py
bsimpson63: This docstring is a big too brief. Can you maybe outline the full process starting from the cron?

reddit_adzerk/adzerkreporting.py
bsimpson63: Sounds reasonable. Mind adding a comment saying that this is the intent? Perhaps in the module docstring.

reddit_adzerk/adzerkreporting.py
bsimpson63: https://github.com/reddit/reddit-public/blob/master/r2/r2/models/traffic.py#L58

I was suggesting this approach mostly to make things more readable by not creating objects outside of the scope where they're going to be used. If you can think of another way to tighten this up that'd be fine as well.

I think the connection is only opened when used, so having it at the module level doesn't result in an open connection.

reddit_adzerk/adzerkreporting.py
bsimpson63: This is another readability comment. It's a lot of mental effort to read through a method that does many things. If this transformation could happen elsewhere that would be better.

reddit_adzerk/adzerkreporting.py
bsimpson63: That's putting a lot of faith in adzerk and that this will only happen when adzerk is still processing. What if this condition remained for a while and we didn't update clicks and impressions for this ad? Instead, could you write out all the data and filter it at display time so that clicks don't exceed impressions?

reddit_adzerk/adzerkreporting.py
bsimpson63: Again, a readability thing. I don't want to have to understand the adzerk report format when I'm reading this.

dwick:
:haircut:

dwick:
:haircut:

reddit_adzerk/adzerkreporting.py
```diff
@@ -0,0 +1,458 @@
...
+    if not date:
+        return None
+
+    return parse_date(date)
+
+
+def _get_fullname(cls, report_fragment):
+    fullname = report_fragment.get("Title", "")
+
+    if not fullname.startswith(cls._fullname_prefix):
```
dwick: the subsequent `_by_fullname` lookup will fail with a less obvious log message without this.  not strictly needed, but i think nice to have

reddit_adzerk/adzerkreporting.py
```diff
@@ -0,0 +1,458 @@
...
+    except report.ReportFailedException as e:
+        g.log.error(e)
+
+        # retry if report failed
+        _generate_promo_report(campaign)
+        return
+
+    impressions, clicks, spent = _get_total_usage(report_result)
+
+    campaign.adserver_spent_pennies = int(spent * 100)
```
dwick: the reporting tables keep this in dollars.  trying to keep the attributes on `PromoCampaign` consistent without having to mess with the traffic stuff (old spent calc is in dollars).

reddit_adzerk/adzerkreporting.py
```diff
@@ -0,0 +1,458 @@
...
+            )
+        else:
+            g.log.warning("adzerk_reporting_q: unknown action - \"%s\"" % action)
+
+    amqp.consume_items("adzerk_reporting_q", _processor, verbose=False)
+
+
+def _insert_daily_link_reporting(
+        codename, date, impressions,
+        clicks, ctr, spent):
```
bsimpson63: `ctr` is never used

reddit_adzerk/adzerkreporting.py
```diff
@@ -0,0 +1,458 @@
...
+        minute=0,
+        second=0,
+        microsecond=0,
+        tzinfo=None,
+    )
+    clicks_row = AdserverClickthroughsByCodename(
+        codename=codename,
+        date=date,
+        interval="day",
+        unique_count=clicks,
```
bsimpson63: since we're not really counting uniques, could you not write this column?

dwick: unfortunately it's required by the traffic code/tables

reddit_adzerk/adzerkreporting.py
bsimpson63: ok

reddit_adzerk/adzerkreporting.py
bsimpson63: ok

reddit_adzerk/adzerkreporting.py
bsimpson63: lame

bsimpson63:
:fish: looks good

reddit_adzerk/adzerkreporting.py
dwick: agreed, cant wait to clean this all up with a new reporting UI

PR #109, opened 2015-11-10
Auction

dwick:
:eyeglasses: @zeantsoi 

made some notes below

### DIFF
--- a/reddit_adzerk/adzerkpromote.py
+++ b/reddit_adzerk/adzerkpromote.py
@@ -53,9 +53,6 @@ from reddit_adzerk.lib.cache import PromoCampaignByFlightIdCache
 hooks = HookRegistrar()
 
 LEADERBOARD_AD_TYPE = 4
-ADZERK_IMPRESSION_BUMP = 500    # add extra impressions to the number we
-                                # request from adzerk in case their count
-                                # is lower than our internal traffic tracking
 
 DELCHARS = ''.join(c for c in map(chr, range(256)) if not (c.isalnum() or c.isspace()))
 
@@ -591,11 +588,11 @@ def delete_campaign(link, campaign):
 
 
 def is_overdelivered(campaign):
-    if not hasattr(campaign, 'cpm') or not campaign.priority.cpm:
+    if campaign.cost_basis != promo.PROMOTE_COST_BASIS.fixed_cpm:
         return False
 
     billable_impressions = promote.get_billable_impressions(campaign)
-    return billable_impressions >= campaign.impressions + ADZERK_IMPRESSION_BUMP
+    return billable_impressions >= campaign.impressions
 
 
 def process_adzerk():


reddit_adzerk/adzerkpromote.py
```diff
@@ -266,22 +279,31 @@ def update_flight(link, campaign, az_campaign):
...
+            'GoalType': GOAL_TYPE_BY_COST_BASIS[campaign.cost_basis], # goal should be based on billing type
+            'RateType': RATE_TYPE_BY_COST_BASIS[campaign.cost_basis], 
         })
 
+        if campaign.cost_basis == promo.PROMOTE_COST_BASIS.fixed_cpm:
+            d['Impressions'] = campaign.impressions
+        else:
+            d.update({
+                'CapType': 4,
+                'DailyCapAmount': campaign.total_budget / campaign.ndays,
```
dwick: assuming `total_budget` is in cents, but i wasn't sure

reddit_adzerk/adzerkpromote.py
```diff
@@ -266,22 +279,31 @@ def update_flight(link, campaign, az_campaign):
...
+            'GoalType': 2, # percentage
+            'RateType': 1, # flat
         })
     else:
         d.update({
-            'Price': campaign.bid,
-            'Impressions': 100,
-            'GoalType': 2, # 2: Percentage
-            'RateType': 1, # 1: Flat
+            'Price': campaign.bid_by_cost_basis / 100., # convert cents to dollars
```
dwick: relies on this change being made https://github.com/reddit/reddit-public/pull/4307#discussion_r44360517

reddit_adzerk/adzerkpromote.py
```diff
@@ -266,22 +279,31 @@ def update_flight(link, campaign, az_campaign):
     else:
         d['IsFreqCap'] = None
 
-    is_cpm = hasattr(campaign, 'cpm') and campaign.priority.cpm
-    if is_cpm:
+    if campaign.priority == promo.HOUSE:
```
dwick: ref: https://github.com/reddit/reddit-public/pull/4307/files#r44362665

reddit_adzerk/adzerkpromote.py
zeantsoi: So I talked with James @ Adzerk about this – I had assumed this should be in cents, but he was insistent that this come through a fixed point float (dollars.cents).

I dunno if they've made any changes, but for now, this all assumes that `total_budget` is something like _$123.45_.

reddit_adzerk/adzerkpromote.py
zeantsoi: If you're staunch about checking for equality rather than accessing an attribute, I'll make the change, but I don't think the argument for doing so is compelling. I think that using an attribute is sensible and simple, and doesn't require access to the `promo` module.

reddit_adzerk/adzerkpromote.py
zeantsoi: I made the change to cents in reddit-public; we need to send to Adzerk in dollars.

dwick:
:haircut:

zeantsoi:
:fish:

dwick:
:haircut: found something using `priority.cpm`

reddit_adzerk/adzerkpromote.py
```diff
@@ -568,11 +588,11 @@ def delete_campaign(link, campaign):
 
 
 def is_overdelivered(campaign):
```
dwick: this should go away all together once i get the reporting stuff out.  reporting won't block auction testing though (is a blocker for public release)

zeantsoi:
LGTM. Ready for fish, or should hold off post-testing?

dwick:
this was accidentally merged and subsequently pushed off. new pr here: https://git.io/v4R3p

PR #114, opened 2015-11-17
Auction attrs

dwick:
:eyeglasses: @zeantsoi https://github.com/reddit/reddit-plugin-adzerk/pull/109

PR #115, opened 2015-11-17
Auction

zeantsoi:
:eyeglasses: @dwick 

This builds off Dave's https://github.com/reddit/reddit-plugin-adzerk/pull/114 PR. It does a few things needed to pass the proper keywords to Adzerk. I'm not sure what the precise implications are of changing these, so we should run them by Adzerk (if you don't know already, Dave):

1. Make it optional to pass `Impressions` to Adzerk
2. Adds `DailyCapAmount`, `LifetimeCapAmount`, and `CapType` to `Flight` defaults
3. Doesn't pass `CapType` for auction (it was previously set to `4`, which returns an error)

Here's a few things to note:

- `CapType` defaults to `Impressions` (i.e., in the case of auctions)
- `GoalAmount` evidently defaults to `100` for auction campaigns
- Adzerk UI looks to be missing entries for fields we're passing in, though they may be unimportant for display


dwick:
https://github.com/adzerk/adzerk-api/wiki/Budget-Capping-(Beta)

`CapType` should be `4` according to those docs.  Adzerk has yet to turn this on for our account as we are suppose to test things without budget caps first.  :+1: on the optional fields stuff.

dwick:
put the changes needed into #114 and rebased.

PR #117, opened 2015-11-19
reporting fixes

dwick:
:eyeglasses: @zeantsoi found some bugs and added some more logging to make life a bit easier.

### DIFF
--- a/reddit_adzerk/__init__.py
+++ b/reddit_adzerk/__init__.py
@@ -19,6 +19,7 @@ class Adzerk(Plugin):
             'az_selfserve_salesperson_id',
             'az_selfserve_mobile_web_site_id',
             'az_selfserve_network_id',
+            'az_reporting_timeout',
         ],
 
         ConfigValue.dict(ConfigValue.str, ConfigValue.int): [
--- a/reddit_adzerk/adzerkreporting.py
+++ b/reddit_adzerk/adzerkreporting.py
@@ -190,7 +190,7 @@ def _handle_generate_daily_link_report(link_id):
     # to redo it.  use the last time the report was run as a 
     # starting point, but subtract 24hrs since initial numbers
     # are preliminary.
-    if getattr(link, "last_daily_report_run") is not None:
+    if hasattr(link, "last_daily_report_run"):
         start = max([
             link.last_daily_report_run - timedelta(hours=24),
             link_start,
--- a/reddit_adzerk/adzerkreporting.py
+++ b/reddit_adzerk/adzerkreporting.py
@@ -177,7 +177,10 @@ def _get_flight_id(report_fragment):
 def _handle_generate_daily_link_report(link_id):
     now = datetime.utcnow()
     link = Link._byID(link_id, data=True)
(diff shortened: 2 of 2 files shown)


reddit_adzerk/__init__.py
```diff
@@ -19,6 +19,7 @@ class Adzerk(Plugin):
             'az_selfserve_salesperson_id',
             'az_selfserve_mobile_web_site_id',
             'az_selfserve_network_id',
+            'az_reporting_timeout',
```
zeantsoi: I don't see this being used anywhere?

zeantsoi: Is it maybe supposed to be `adserver_reporting_timeout`?

dwick: yes, ty

zeantsoi: Nit: I wonder whether it's obvious that the value here will be passed as _minutes_. Intuitively, I kinda think of the term _timeout_ to be in seconds. Would it be useful to explicitly call that out in the name?

reddit_adzerk/adzerkreporting.py
```diff
@@ -190,7 +197,7 @@ def _handle_generate_daily_link_report(link_id):
     # to redo it.  use the last time the report was run as a 
     # starting point, but subtract 24hrs since initial numbers
     # are preliminary.
-    if getattr(link, "last_daily_report_run") is not None:
+    if getattr(link, "last_daily_report_run", None) is not None:
```
zeantsoi: Does it make sense to check `hasattr(link, 'last_daily_report_run')`? I'm pretty sure the the use would be the same as `getattr` in this instace. I tend towards checking `hasattr`, though, since it's less verbose.

reddit_adzerk/adzerkreporting.py
```diff
@@ -177,7 +181,10 @@ def _get_flight_id(report_fragment):
 def _handle_generate_daily_link_report(link_id):
     now = datetime.utcnow()
     link = Link._byID(link_id, data=True)
-    campaigns = PromoCampaign._by_link(link._id)
+    campaigns = list(PromoCampaign._by_link(link._id))
```
zeantsoi: [Check this out](https://github.com/reddit/reddit-public/blob/master/r2/r2/lib/db/thing.py#L1036). The current `_count` implementation is broken (see my PR [here](https://github.com/reddit/reddit-public/pull/4451)), but I'd imagine that it's more effective to check `PromoCampaign._by_link(link._id)._cursor().rowcount` than generating the list.

zeantsoi:
:nail_care:

reddit_adzerk/adzerkreporting.py
zeantsoi: Nevermind ^. We need to have the `campaigns` iterator anyways.

reddit_adzerk/__init__.py
dwick: changed it to seconds for python consistency.

dwick:
:haircut:

zeantsoi:
:fish:

PR #122, opened 2015-12-08
Display ads: Optionally display a static image in lieu of an ad.

dwick:
This is used to reduce the number of requests made to adzerk on
low fill targets.

:eyeglasses: @zeantsoi 

no config quite yet. it will look something like this:

```
display_ad_skip_probability = .5
display_ad_skip_keywords = nfl, sailing
display_ad_skip_images = https://s.zkcdn.net/Advertisers/f7fb55f0ded043328386e048aa8b06b0.png, https://static.adzerk.net/Advertisers/c602fa14f4f143d1921ac115c5d49ba9.png
```


reddit_adzerk/public/static/js/adzerk/display.js
```diff
@@ -33,6 +33,35 @@
...
+          skipAd = true;
+          break;
+        }
+      }
+    }
+
+    if (skipAd) {
+      var adframe = document.getElementById('main');
+      var img = document.createElement('img');
+      var randomImgIndex = Math.floor(Math.random() * global.SKIP_AD_IMAGES.length) + 0;
```
zeantsoi: What does `+ 0` do?

reddit_adzerk/public/static/js/adzerk/display.js
```diff
@@ -33,6 +33,35 @@
 
   var config = getConfig();
 
+  // Display a random image in lieu of an ad for certain keywords.
+  // This reduces the number of ad requests for low-fill targets.
+  if (global.SKIP_AD_PROBABILITY && Math.random() <= global.SKIP_AD_PROBABILITY) {
+    var keywords = config.keywords ? config.keywords.split(',') : [];
+    var skipAd = !global.SKIP_AD_KEYWORDS;
+
+    if (!skipAd && global.SKIP_AD_KEYWORDS && keywords) {
```
zeantsoi: Do we need to evaluate both `!skipAd` _and_ `global.SKIP_AD_KEYWORDS`? Shouldn't they evaluate the same?

zeantsoi:
I get what we're trying to do here, but it seems like our config will get a bit unwieldy and difficult to maintain. Also, it seems unfitting that we should define static asset locations in the config, but I could be mistaken.

Do we know what impact this will have on click/impression tracking, if at all?

:nail_care:

reddit_adzerk/public/static/js/adzerk/display.js
dwick: ha, nothing.  `Math.floor(Math.random() * max) + min` is a random number between `min` and `max`.  blindly followed the formula...

dwick:
the idea here is to not serve any ads so that we can reduce adzerk usage. the plan is just to have a bunch of "thanks for not using adblock" images to randomly display when we dont show an ad.  we dont care about impression tracking on these since they arent really ads.  the reason for using config is so they can change the images without having to make code changes.  if you have other ideas of how to do that i'm all ears. (maybe a wiki page?)

this will have an effect on adzerk forecasting, but i've already explained this to chris.

dwick:
:haircut:

zeantsoi:
Awaiting config...

dwick:
:haircut: hard coded the images since they're not likely to change very often.

config: https://github.com/reddit/reddit-private/pull/1158

zeantsoi:
Would still love to get another dev's thoughts for a sanity check on this approach. Do we need to get this out urgently?

zeantsoi:
LGTM :fish: 

Might want to indent [this](https://github.com/reddit/reddit-plugin-adzerk/pull/122/files#diff-367dcddf992dcbdbc99e0a2d216db236R4) to two spaces to conform to [convention here](https://github.com/reddit/reddit-plugin-adzerk/blob/master/reddit_adzerk/templates/baseadframe.html#L4).

dwick:
21dc196d90cb9738c0f8954ed1884d2bfc0a1ec5

PR #123, opened 2015-12-09
Display ads: Use `promote.keywords_from_context` to generate keywords

dwick:
:eyeglasses: @bsimpson63 

This was the change that caused memoize to fail, already been reviewed for the code itself, just wanted to make sure nothing else unforeseen happens.

related: https://github.com/reddit/reddit-public/pull/4478

### DIFF
--- a/reddit_adzerk/adzerkads.py
+++ b/reddit_adzerk/adzerkads.py
@@ -14,7 +14,9 @@ from r2.controllers.reddit_base import (
 from r2.lib import promote
 from r2.lib.pages import Ads as BaseAds
 from r2.lib.wrapped import Templated
-from r2.models.subreddit import DefaultSR
+from r2.models.subreddit import (
+    Frontpage,
+)
 
 FRONTPAGE_NAME = "-reddit.com"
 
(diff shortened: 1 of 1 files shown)


reddit_adzerk/adzerkads.py
```diff
@@ -22,18 +24,17 @@ class Ads(BaseAds):
...
+            include_subscriptions=False,
+        )
 
         # adzerk reporting is easier when not using a space in the tag
-        if isinstance(c.site, DefaultSR):
-            site_name = FRONTPAGE_NAME
+        if Frontpage.name in keywords:
+            keywords = keywords - {Frontpage.name} | {FRONTPAGE_NAME}
 
-        sr = quote(site_name.lower())
```
bsimpson63: looks like you're no longer lowercasing all the subreddit names

bsimpson63:
:nail_care: 

Seems ok otherwise.

Now that the lookup is stalecached we should almost never be hitting the "real" cache. It might be nice to have a mechanism for local-only memoization/caching for cases like this.

Did you consider adding an attribute to the `Subreddit` object indicating whether it was nsfw for advertising purposes? That seems like a more direct way to do what you want and wouldn't be too big of a pain because collections don't change often.

reddit_adzerk/adzerkads.py
dwick: this wasn't needed: http://help.adzerk.com/hc/en-us/articles/202624319-Keyword-Targeting

dwick:
> Did you consider adding an attribute to the Subreddit object indicating whether it was nsfw for advertising purposes

This is something we want to do on a larger scale eventually. "ads categories".. I'd rather not have to do it right now though

:haircut:

bsimpson63:
:cry: ok

:fish:

PR #126, opened 2015-12-10
Use frequency_cap_duration_hours from reddit-public config

zeantsoi:
:eyeglasses: @dwick 

Per https://github.com/reddit/reddit-public/pull/4489.


dwick:
is there any reason to not just hard code this to 24hrs? we don't expose this in the UI from what I can tell.

:nail_care:

zeantsoi:
I don't have a _really_ compelling argument for why it must be configured this way, but I think it's more sensible to keep this stuff in the config for the following two reasons:

1. It's something that _can_ be dialed up/down, and it seem to be the general pattern to keep such configurable options in the config;
2. The request to provide dynamic duration originally came from Chris K., but Zubair wanted it taken out. Given how much flux they seem to be in over issues like this, it seems wise to leave things maximally configurable.

But if you have a better argument for hardcoding, I'm happy to entertain.

:haircut:

dwick:
@zeantsoi does the time period show up in the UI? if not i'd say hard code it.

dwick:
it wouldn't make much sense to me to have it be 3 for every 17 hours or something like that.  I just don't see it being anything other than 24.

zeantsoi:
Shows up [here](https://github.com/reddit/reddit-public/pull/4489/files#diff-9fe1c42555d19eedc61078ca51ac62dcR594) and [here](https://github.com/reddit/reddit-public/pull/4489/files#diff-9fe1c42555d19eedc61078ca51ac62dcR608).

I could think of a few intervals that make sense – one hour, 48, 72, etc. The purpose here is flexibility – after all, it _was_ ad ops that first requested that we make duration configurable in the first place.

zeantsoi:
Hardcoded frequency cap duration :haircut:

dwick:
:fish:

zeantsoi:
af674cbf9c5250f009ad7e96301540666f893e7c

PR #128, opened 2015-12-18
Fix bug with ads being skipped when they should not be

dwick:
`keywords` is an array and needs to be iterated as such.

:eyeglasses: @zeantsoi 

before:
screen shot 2015-12-18 at 1 13 45 pm
after:
screen shot 2015-12-18 at 1 15 30 pm


zeantsoi:
:fish:

PR #132, opened 2015-12-28
Ad reporting sync

dwick:
:eyeglasses: @zeantsoi


dwick:
:eyeglasses: @bsimpson63 i think z is still on vacation

dwick:
1 second

PR #134, opened 2016-01-05
Reporting: Ensure proper start date is used

dwick:
Fixes an edge case when back filling data.  If a report is run a
2nd time well after the campaign has already completed then the
start date used would be that of the first run of the report.
Which in this case is usually greater than the end date.

:eyeglasses: @zeantsoi


zeantsoi:
:fish:

PR #136, opened 2016-01-08
Instrument

dwick:
:eyeglasses: @zeantsoi @Deimos 

ticket: https://reddit.atlassian.net/browse/ADS-256
spec: https://reddit.atlassian.net/wiki/display/DA/Ad+Serving+Events

cc @drunken-economist

### DIFF
--- a/reddit_adzerk/__init__.py
+++ b/reddit_adzerk/__init__.py
@@ -1,6 +1,8 @@
 import json
 import pkg_resources
 
+from pylons import app_globals as g
+
 from r2.lib.plugin import Plugin
 from r2.lib.configparse import ConfigValue
 from r2.lib.js import Module
@@ -22,6 +24,7 @@ class Adzerk(Plugin):
 
         ConfigValue.float: [
             'display_ad_skip_probability',
+            'events_collector_ad_serving_sample_rate',
         ],
 
         ConfigValue.tuple: [
@@ -72,7 +75,11 @@ class Adzerk(Plugin):
         # replace the standard Ads view with an Adzerk specific one.
         import r2.lib.pages.pages
         from adzerkads import Ads as AdzerkAds
+        from lib.events import AdEventQueue
+
         r2.lib.pages.pages.Ads = AdzerkAds
+
+        g.ad_events = AdEventQueue()
 
         # replace standard adserver with Adzerk.
         from adzerkpromote import AdzerkApiController
--- a/reddit_adzerk/adzerkpromote.py
+++ b/reddit_adzerk/adzerkpromote.py
@@ -661,7 +661,7 @@ class AdserverResponse(object):
 
 
 def adzerk_request(keywords, uid, num_placements=1, timeout=1.5,
-                   platform="desktop"):
+                   platform="desktop", is_refresh=False):
     placements = []
     divs = ["div%s" % i for i in xrange(num_placements)]
 
@@ -675,9 +675,10 @@ def adzerk_request(keywords, uid, num_placements=1, timeout=1.5,
         }
         placements.append(placement)
 
+    keywords = [word.lower() for word in keywords]
     data = {
         "placements": placements,
-        "keywords": [word.lower() for word in keywords],
+        "keywords": keywords,
         "ip": request.ip,
     }
 
@@ -702,6 +703,18 @@ def adzerk_request(keywords, uid, num_placements=1, timeout=1.5,
 
     timer = g.stats.get_timer("providers.adzerk")
     timer.start()
+
+    for placement in placements:
+        g.ad_events.ad_request(
+            keywords=keywords,
+            platform=platform,
+            placement_name=placement["divName"],
+            placement_types=placement["adTypes"],
+            is_refresh=is_refresh,
+            subreddit=c.site,
+            request=request,
+            context=c,
+        )
 
     try:
         r = requests.post(url, data=json.dumps(data), headers=headers,
@@ -728,17 +741,33 @@ def adzerk_request(keywords, uid, num_placements=1, timeout=1.5,
     if not decisions:
         return None
 
+    placements_by_div = {placement["divName"]: placement
+        for placement in placements}
+
     res = []
     for div in divs:
         decision = decisions[div]
         if not decision:
             continue
 
+        placement = placements_by_div[div]
+        ad_id = decision['adId']
+
         # adserver ads are not reddit links, we return the body
         if decision['campaignId'] in g.adserver_campaign_ids:
+            g.ad_events.ad_response(
+                keywords=keywords,
+                platform=platform,
+                placement_name=div,
+                placement_types=placement["adTypes"],
+                ad_id=ad_id,
+                subreddit=c.site,
+                request=request,
+                context=c,
+            )
+
             return AdserverResponse(decision['contents'][0]['body'])
 
-        adzerk_campaign_id = decision['campaignId']
         adzerk_flight_id = decision['flightId']
         imp_pixel = decision['impressionUrl']
         click_url = decision['clickUrl']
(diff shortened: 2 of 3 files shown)


drunken-economist:
`event_topi`c for these events should be `ad_events`, or even something more descriptive. This would keep it consistent with other event naming conventions. The basic idea is to name the event topic specifically enough that its contained `event_types` can be understood with exploring the data. If you want all events for ads in the same umbrella, `ad_events` is fine, otherwise, a logical partition might be needed

dwick:
@drunken-economist how about `ad_serving` or `ad_serving_events`?

drunken-economist:
@dwick `ad_serving_events` sounds good to me (all values for `event_topic` should end in `_events` to keep consistency)

reddit_adzerk/__init__.py
```diff
@@ -22,6 +24,7 @@ class Adzerk(Plugin):
 
         ConfigValue.float: [
             'display_ad_skip_probability',
+            'events_collector_ad_serving_sample_rate',
```
zeantsoi: Nit: this is kinda verbose. Whaddya think about just `ad_service_sample_rate`?

reddit_adzerk/adzerkpromote.py
```diff
@@ -727,17 +739,34 @@ def adzerk_request(keywords, uid, num_placements=1, timeout=1.5,
     if not decisions:
         return None
 
+    placements_by_div = {
```
zeantsoi: Nit: this indentation here reads strangely to me. How about:

    placements_by_div = {placement["divName"]: placement
    	for placement in placements}

reddit_adzerk/lib/events.py
```diff
@@ -0,0 +1,122 @@
+from pylons import app_globals as g
```
zeantsoi: I once got a comment that we should be adding our license header to files, though I'm not sure whether that uniformly applies to plugins. However, seeing as `reddit-plugin-adzerk` is public, we should probably add it to new files here.

reddit_adzerk/__init__.py
```diff
@@ -72,8 +75,12 @@ def load_controllers(self):
         # replace the standard Ads view with an Adzerk specific one.
         import r2.lib.pages.pages
         from adzerkads import Ads as AdzerkAds
+        from lib.events import AdEventQueue
+
         r2.lib.pages.pages.Ads = AdzerkAds
 
+        g.ad_events = AdEventQueue()
```
zeantsoi: What's the purpose of making this global?

zeantsoi:
:nail_care: this is a great start for instrumenting. Just a couple of non-blocking comments.

reddit_adzerk/__init__.py
dwick: this follows the format of the other sample rates: https://github.com/reddit/reddit-private/blob/af17559429aeaf222a685d0817ec18aeede5bd55/production.ini#L935-L950

reddit_adzerk/__init__.py
dwick: it's a singleton that might need to be accessed from anywhere.  the standard events queue is `g.events` as well.

reddit_adzerk/lib/events.py
dwick: literally nothing in this project has a license header so i'm reluctant to go adding them.  there's also this: https://github.com/reddit/reddit-plugin-adzerk/blob/master/LICENSE, which i think should cover it. probably only an issue for reddit-public.

dwick:
:haircut:

reddit_adzerk/lib/events.py
```diff
@@ -0,0 +1,122 @@
...
+    EventQueue,
+    Event,
+    squelch_exceptions,
+)
+from r2.models import (
+    FakeSubreddit,
+)
+
+def _is_subreddit(subreddit):
+    return subreddit and not isinstance(subreddit, FakeSubreddit)
```
Deimos: Minor nit, but this should probably be:

    return bool(subreddit and not isinstance(subreddit, FakeSubreddit))

Otherwise it's going to be a little weird that it will return the subreddit when true, but boolean `False` otherwise.

reddit_adzerk/lib/events.py
```diff
@@ -0,0 +1,122 @@
...
+            context=context,
+        )
+
+        event.add("keywords", keywords)
+        event.add("platform", platform)
+        event.add("placement_name", placement_name)
+        event.add("placement_types", placement_types)
+
+        if _is_subreddit(subreddit):
+            event.add("sr_id", subreddit._id)
```
Deimos: You can use `event.add_subreddit_fields(subreddit)` here instead, which will add both `sr_id` and `sr_name`.

And since that function does a check for whether `subreddit` is true itself, it might be overall simpler to just drop the `_is_subreddit` function and replace this with:

    if not isinstance(subreddit, FakeSubreddit):
        event.add_subreddit_fields(subreddit)

reddit_adzerk/lib/events.py
```diff
@@ -0,0 +1,122 @@
...
+            event.add("sr_name", subreddit.name)
+
+        self.save_event(event)
+
+    @squelch_exceptions
+    @sampled("events_collector_ad_serving_sample_rate")
```
Deimos: Just to confirm: do you want to have both of these events tied to the same sampling rate, or would it be better to give them individual ones?

reddit_adzerk/lib/events.py
```diff
@@ -0,0 +1,122 @@
...
+            placement_name,
+            placement_types,
+            subreddit=None,
+            request=None,
+            context=None,
+        ):
+        """Create an `ad_request` for event-collector.
+
+        keywords: Array of keywords used to select the ad.
+        platform: the platform the ad was request for.
```
Deimos: Super super minor nit, but this line's grammar is kinda off, and it's also the only one that isn't capitalized.

reddit_adzerk/lib/events.py
```diff
@@ -0,0 +1,122 @@
...
+            placement_name,
+            placement_types,
+            ad_id,
+            link_id=None,
+            campaign_id=None,
+            subreddit=None,
+            request=None,
+            context=None,
+        ):
+        """Create an `ad_request` for event-collector.
```
Deimos: This is incorrect, should say `ad_response`

reddit_adzerk/lib/events.py
```diff
@@ -0,0 +1,122 @@
...
+
+        keywords: Array of keywords used to select the ad.
+        platform: the platform the ad was request for.
+        placement_name: The identifier of the placement.
+        placement_types: Array of placements types.
+        ad_id: Unique id of the ad response.
```
Deimos: Missing a line in here describing the `link_id` argument.

reddit_adzerk/lib/events.py
```diff
@@ -0,0 +1,122 @@
...
+            context=context,
+        )
+
+        event.add("keywords", keywords)
+        event.add("platform", platform)
+        event.add("placement_name", placement_name)
+        event.add("placement_types", placement_types)
+        event.add("ad_id", ad_id)
+
+        if link_id:
```
Deimos: You don't need to check for truthiness any more, `event.add()` handles that itself, so you can just try to add `link_id` and it will only happen if it has a value.

reddit_adzerk/lib/events.py
```diff
@@ -0,0 +1,122 @@
...
+            event.add("link_id", link_id)
+
+        if campaign_id:
```
Deimos: Same note here as above about the truthiness check being unnecessary.

reddit_adzerk/lib/events.py
```diff
@@ -0,0 +1,122 @@
...
+            event.add("campaign_id", campaign_id)
+
+        if _is_subreddit(subreddit):
```
Deimos: Same notes as from `ad_request` for using `add_subreddit_fields()` and/or switching the check.

Deimos:
Just minor stuff, overall looks good. :nail_care:

reddit_adzerk/lib/events.py
dwick: :+1:

reddit_adzerk/lib/events.py
zeantsoi: @Deimos any thoughts? ^

This _is_ a public repo, BTW.

Deimos: Hmm, I'm not sure (and not really qualified to make that call). [The gold plugin repo](https://github.com/reddit/reddit-plugin-gold) is also public and doesn't have licenses in each file either.

reddit_adzerk/lib/events.py
dwick: i think it makes sense to use the same sample rate since they're highly related.

dwick:
:haircut:

reddit_adzerk/lib/events.py
```diff
@@ -0,0 +1,117 @@
...
+from r2.lib.eventcollector import (
+    EventQueue,
+    Event,
+    squelch_exceptions,
+)
+from r2.models import (
+    FakeSubreddit,
+)
+
+def _is_subreddit(subreddit):
```
Deimos: You can remove this function now, since it's not being used any more.

Deimos:
Looks good, just need to drop the now-unnecessary `_is_subreddit` function. :fish: outside of that.

dwick:
whoops, got that and rebased.

drunken-economist:
:fish: lgtm

PR #137, opened 2016-01-08
Use the same keywords as sponsored headlines

dwick:
Headlines now use a system keyword (s.frontpage) for the frontpage.
This keeps things consistent.

:eyeglasses: @zeantsoi 

We'll need to either write a script or get adzerk to migrate all the keywords.


zeantsoi:
:fish:

dwick:
059ae242537df1491943a878c7cff9f4afd1d46a

PR #139, opened 2016-01-14
Fix `events_collector_ad_serving_sample_rate` type specification

dwick:
:eyeglasses: @Deimos doh!


Deimos:
Ah, that makes sense. :fish:

PR #140, opened 2016-01-15
Remove delay on flight's start time

dwick:
:eyeglasses: @bsimpson63 

request from ad ops. roadblocked campaigns need to start on time, figured this probably isn't necessary anymore.


bsimpson63:
This was requested by adzerk https://github.com/reddit/reddit-plugin-adzerk/pull/30

I think the issue was that they start trying to show the ad before it actually launches so that they can calibrate it for when it actually launches, but that wasn't working because we don't allow an ad to be shown if it's not live on our end.

dwick:
i guess if we supported roadblocks from our UI then we could remove the delay since they don't require balancing? checking in with james on this.

bsimpson63:
:running: but :eyeglasses:  me if you want to pick this up again

PR #141, opened 2016-01-21
Use site to derive keywords directly if provided

dwick:
This allows clients to not have to worry about the logic around
which keywords (srnames) to send.

:eyeglasses: @zeantsoi

### DIFF
--- a/reddit_adzerk/adzerkpromote.py
+++ b/reddit_adzerk/adzerkpromote.py
@@ -22,8 +22,13 @@ from r2.lib import (
     organic,
     promote,
 )
+from r2.lib.base import abort
 from r2.lib.csrf import csrf_exempt
 from r2.lib.db.sorts import epoch_seconds
+from r2.lib.errors import (
+    errors,
+    reddit_http_error,
+)
 from r2.lib.filters import _force_utf8
 from r2.lib.pages import responsive
 from r2.lib.pages.things import default_thing_wrapper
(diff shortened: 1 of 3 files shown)


reddit_adzerk/adzerkpromote.py
```diff
@@ -833,21 +842,28 @@ class AdzerkApiController(api.ApiController):
         loid=nop('loid', None),
         is_refresh=VBoolean("is_refresh", default=False),
     )
-    def POST_request_promo(self, srnames, is_mobile_web, platform, loid, is_refresh):
+    def POST_request_promo(self, site, srnames, is_mobile_web, platform, loid, is_refresh):
         self.OPTIONS_request_promo()
 
-        if not srnames:
+        if (errors.INVALID_SITE_PATH, "site") in c.errors:
+            return abort(reddit_http_error(400, errors.INVALID_SITE_PATH))
```
zeantsoi: A lot of other errors in this method are handled simply by making a `return`. Is it necessary to explicitly return a 400? Are we deliberately trying to make the failures loud?

reddit_adzerk/lib/validator.py
```diff
@@ -0,0 +1,31 @@
+import re
+
+from r2.lib.errors import errors
+from r2.lib.validator import (
+    sr_path_rx,
```
zeantsoi: Unused import.

zeantsoi:
:nail_care:

reddit_adzerk/adzerkpromote.py
dwick: i just want to make this obvious for whoever is using it.  i'd consider it a 400 because someone was supplied, but was invalid.

dwick:
:haircut:

zeantsoi:
:fish:

PR #144, opened 2016-01-28
Force ndays to be at least 1 to avoid dividing by zero

zeantsoi:

[Brian's observation](https://github.com/reddit/reddit-public/pull/4685#issuecomment-176245155) is correct, and is actually precisely what caused the exception (i.e., campaign terminated on start date). This solution is less heavy-handed and does the minimum to prevent an exception from being thrown for bad math.


bsimpson63:
:fish:

dwick:
:fish:

zeantsoi:
8e16bf378e5fce54377e518c4e4799fac7bcd148

PR #145, opened 2016-02-01
Fix import error

dwick:
FRONTPAGE_NAME is now only used for legacy reporting.  Moved it to
the report module.

:eyeglasses: @bsimpson63


bsimpson63:
:fish:

PR #149, opened 2016-02-09
Stop unnecessarily updating `SiteZoneTargeting`

dwick:
SiteZoneTargeting hardly ever changes, but causes unchanged flights
to be updated unnecessarily/spam the `PromotionLog` because adzerk
also includes `Id`s in the GET results.

:eyeglasses: @zeantsoi

### DIFF
--- a/reddit_adzerk/adzerkpromote.py
+++ b/reddit_adzerk/adzerkpromote.py
@@ -381,12 +381,28 @@ def update_flight(link, campaign):
         })
 
     if campaign.platform != 'all':
-        d.update({
-            'SiteZoneTargeting': [{
-                'SiteId': g.az_selfserve_site_ids[campaign.platform],
-                'IsExclude': False,
-            }],
-        })
+        site_targeting = [{
+            'SiteId': g.az_selfserve_site_ids[campaign.platform],
+            'IsExclude': False,
+            'ZoneId': None,
+            'FlightId': az_flight.Id if az_flight else None,
+        }]
+        # Check equality more specifically to reduce spam in the PromotionLog.
+        update_site_targeting = True
+        if az_flight and az_flight.SiteZoneTargeting:
+            # Loop through the existing site targeting and remove the `Id` param.
+            old_site_targeting = map(
+                lambda (index, value): {
+                    key: value[key]
+                    for key in value.keys() if key != "Id"},
+                enumerate(az_flight.SiteZoneTargeting),
+            )
+            update_site_targeting = old_site_targeting != site_targeting
+
+        if update_site_targeting:
+            d.update({
+                'SiteZoneTargeting': site_targeting,
+            })
 
     # special handling for location conversions between reddit and adzerk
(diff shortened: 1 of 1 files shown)


reddit_adzerk/adzerkpromote.py
```diff
@@ -381,12 +381,27 @@ def update_flight(link, campaign, az_campaign):
...
-        })
+        site_targeting = [{
+            'SiteId': g.az_selfserve_site_ids[campaign.platform],
+            'IsExclude': False,
+            'ZoneId': None,
+            'FlightId': az_flight.Id if az_flight else None,
+        }]
+        # Check equality more specifically to reduce spam in the PromotionLog.
+        update_site_targeting = True
+        if (az_flight and az_flight.SiteZoneTargeting):
```
zeantsoi: Nit: parentheses not required.

reddit_adzerk/adzerkpromote.py
```diff
@@ -381,12 +381,27 @@ def update_flight(link, campaign, az_campaign):
...
+            old_site_targeting = map(
```
zeantsoi: This is kinda hairy to follow. Would an if statement work equally as well? If so, I think it'd read better.

reddit_adzerk/adzerkpromote.py
```diff
@@ -381,12 +381,27 @@ def update_flight(link, campaign, az_campaign):
...
-        d.update({
-            'SiteZoneTargeting': [{
-                'SiteId': g.az_selfserve_site_ids[campaign.platform],
-                'IsExclude': False,
-            }],
-        })
+        site_targeting = [{
+            'SiteId': g.az_selfserve_site_ids[campaign.platform],
+            'IsExclude': False,
+            'ZoneId': None,
```
zeantsoi: Why are we setting ZoneId and FlightId now, but not before?

zeantsoi:
:nail_care:

reddit_adzerk/adzerkpromote.py
dwick: these were setup automatically before, just ensuring we dont have to diff against them.

reddit_adzerk/adzerkpromote.py
dwick: not sure how you would write this otherwise, but i added a comment about what it's doing.

dwick:
:haircut: low priority

reddit_adzerk/adzerkpromote.py
zeantsoi: :no_bell: 

Something like this would work:

	for flight in az_flight.SiteZoneTargeting:
	    for key in flight.keys():
	            if key == 'Id':
	                    del event[key]

zeantsoi:
Optional observation [here](https://github.com/reddit/reddit-plugin-adzerk/pull/149#discussion_r52982718), :fish: otherwise.

PR #150, opened 2016-02-10
Fix DNT for sponsored headlines

dwick:
:eyeglasses: @zeantsoi


zeantsoi:
:fish: 

Weird... I recall having [done this](https://github.com/reddit/reddit-public/blob/master/r2/r2/lib/promote.py#L264-L267) to avoid a naming collision... but I don't see where it would have collided.

PR #154, opened 2016-02-24
Move adzerk ad controls so they do not overlap adchoices

dwick:
:eyeglasses: @zeantsoi 

screen shot 2016-02-24 at 2 46 07 pm


zeantsoi:
:fish:

dwick:
the location of the adchoice banner depends on the advertiser :\, going to have to rework this a bit.

PR #156, opened 2016-03-02
update_adzerk: Only update flight data when a campaign is provided.

dwick:
Previously we were updating everything about the campaign,
advertiser, and creative when doing simple actions like pausing a
flight.  This should reduce the cost of updates.

:eyeglasses: @zeantsoi

### DIFF
--- a/reddit_adzerk/adzerkpromote.py
+++ b/reddit_adzerk/adzerkpromote.py
@@ -266,7 +266,7 @@ def update_advertiser(author):
     return az_advertiser
 
 
-def update_flight(link, campaign, az_campaign):
+def update_flight(link, campaign):
     """Add/update a reddit campaign as an Adzerk Flight"""
     if getattr(campaign, 'external_flight_id', None) is not None:
         az_flight = adzerk_api.Flight.get(campaign.external_flight_id)
@@ -304,7 +304,7 @@ def update_flight(link, campaign, az_campaign):
         'IsUnlimited': False,
         'IsFullSpeed': False,
         'Keywords': '\n'.join(list(keywords)),
-        'CampaignId': az_campaign.Id,
+        'CampaignId': link.external_campaign_id,
         'PriorityId': g.az_selfserve_priorities[campaign.priority_name],
         'IsDeleted': False,
         'IsActive': (not promote.campaign_needs_approval(link, campaign) and
@@ -482,7 +482,7 @@ def update_flight(link, campaign, az_campaign):
     return az_flight
 
 
-def create_cfmap(link, campaign, az_campaign, az_creative, az_flight):
+def create_cfmap(link, campaign):
     """Create a CreativeFlightMap.
 
     Map the the reddit link (adzerk Creative) and reddit campaign (adzerk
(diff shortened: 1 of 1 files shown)


reddit_adzerk/adzerkpromote.py
```diff
@@ -482,7 +482,7 @@ def update_flight(link, campaign, az_campaign):
     return az_flight
 
 
-def create_cfmap(link, campaign, az_campaign, az_creative, az_flight):
+def create_cfmap(link, campaign):
```
zeantsoi: `campaign` should be an optional defaulted to `None`, I think?

reddit_adzerk/adzerkpromote.py
```diff
@@ -540,22 +540,19 @@ def _update_adzerk(link, campaign):
...
-        az_advertiser = update_advertiser(author)
-        az_campaign = update_campaign(link, az_advertiser)
-        az_creative = update_creative(link, az_advertiser)
 
-        if campaign:
-            az_flight = update_flight(link, campaign, az_campaign)
+        if campaign is None:
+            author = Account._byID(link.author_id, data=True)
+            az_advertiser = update_advertiser(author)
+            az_campaign = update_campaign(link, az_advertiser)
```
zeantsoi: Is this needed?

reddit_adzerk/adzerkpromote.py
zeantsoi: Or, if not, where is `None` being assigned to the campaign being passed in? Separate commit?

dwick: campaign is required here

reddit_adzerk/adzerkpromote.py
dwick: what needed?

reddit_adzerk/adzerkpromote.py
zeantsoi: In what instance would it be `None`?

reddit_adzerk/adzerkpromote.py
zeantsoi: `update_campaign`.

zeantsoi: Nevermimd, mixed up the nomenclature.

reddit_adzerk/adzerkpromote.py
dwick: i think you mean for `update_adzerk`? https://github.com/reddit/reddit-plugin-adzerk/blob/30d5548253b3c79c11371592f3412055497e4841/reddit_adzerk/adzerkpromote.py#L609

reddit_adzerk/adzerkpromote.py
dwick: reddit link == adzerk campaign and creative
reddit campaign == adzerk flight and map

reddit_adzerk/adzerkpromote.py
zeantsoi: Whoops, you're right.

reddit_adzerk/adzerkpromote.py
zeantsoi: Just an idea, but being that we'll probably be adding some people over the next few months, maybe we should add some inline comments to this end.

zeantsoi:
:fish:

PR #157, opened 2016-03-16
Ensure that GroupBy passes a value

zeantsoi:
:eyeglasses: @kjoconnor 

cc: @dwick 

Per Adzerk, we need to pass a value, otherwise we get a 400 response: `{"message":"Argument cannot be null.\nParameter name: source"}`.

### DIFF
--- a/reddit_adzerk/report.py
+++ b/reddit_adzerk/report.py
@@ -65,6 +65,9 @@ def demangle_frontpage_name(keyword):
 
 
 def queue_report(start, end, groups=None, parameters=None):
+    if not groups:
+        groups = []
+
     data = {
         "StartDate": start.strftime("%m/%d/%Y"),
         "EndDate": end.strftime("%m/%d/%Y"),


reddit_adzerk/report.py
```diff
@@ -64,7 +64,7 @@ def demangle_frontpage_name(keyword):
         return keyword
 
 
-def queue_report(start, end, groups=None, parameters=None):
+def queue_report(start, end, groups=[], parameters=None):
```
kjoconnor: This probably won't work the way you think it might, it's not a good idea to use mutables as default arguments weirdly enough!

Some info here: https://pythonconquerstheuniverse.wordpress.com/2012/02/15/mutable-default-arguments/

I'd instead leave this as `groups=None`, and then in the first line of the function do something like:
```python
if groups is None:
    groups = []
```

kjoconnor:
:nail_care:

reddit_adzerk/report.py
zeantsoi: Good call! Haven't thought about that for ages.

And to think: my PRs are usually flawless!

zeantsoi:
Squashed and force pushed :haircut:

reddit_adzerk/report.py
```diff
@@ -65,6 +65,9 @@ def demangle_frontpage_name(keyword):
 
 
 def queue_report(start, end, groups=None, parameters=None):
+    if not groups:
```
kjoconnor: Do we want this to be `not groups` or are we specifically checking for `None`? If we pass `False` for groups do we want that to be passed along to Adzerk?

kjoconnor:
:nail_care:

reddit_adzerk/report.py
zeantsoi: We necessarily must send an array, whether empty or not. It's okay to check for falsy here.

zeantsoi:
:haircut:

reddit_adzerk/report.py
kjoconnor: This may be overkill, but if it *has* to be an array then you can also check `isinstance(list, groups)` and cast it to an empty list instead. Then if someone provides `4` or `iloveadzerk` it will not let that garbage be passed through.

It could also be the case that you want it to error out if somehow someone is calling this function without providing properly typed arguments. I'll leave it up to you guys since you know how best this stuff is used.

kjoconnor:
:fish: a comment about possibly casting to a list for bad types, but fine regardless of what you end up with there. Thanks for fixing!

PR #158, opened 2016-03-27
Send request payload to event collector on flight creation/update

zeantsoi:
:eyeglasses: @dwick @drunken-economist 

cc: @kjoconnor 

[Proposed event spec here](https://reddit.atlassian.net/wiki/pages/viewpage.action?pageId=48005166).

We're having a bunch of inconsistencies between our data and Adzerk's data. Per discussion with Kavin and James (Adzerk), we need to start logging outgoing requests for Adzerk object creation, starting with flights. It should be sufficient to log the request payload and object identifier.

Per discussion with @kjoconnor, it sounds like the best way to log the outgoing request payload is sending an event, rather than writing to a logfile.

@drunken-economist, FYI: this purpose of this event is more for auditing, rather than statistical record-keeping. Not sure if this is definitively the best place to be doing this, but it would be helpful to be able to run reports via Interana, rather than grepping logfiles.


reddit_adzerk/lib/events.py
```diff
@@ -113,3 +113,34 @@ def ad_response(
...
+        """
+        if request_type not in ('create', 'update'):
+            return
+
+        event = AdEvent(
+            topic='ad_upsert_events',
+            event_type='ss.ad_%s_request' % request_type,
+        )
+
+        event.add('ad_object_fullname', ad_object_fullname)
```
dwick: i think we probably want to be consistent with other events and call this `thing_fullname`

dwick:
i could see wanting to add some more data to this.  advertiser name/id could be using for finding problems related to certain advertisers.  also, might it be possible to add the user name/id for the person who triggered the update for ops/vs advertiser changes?

:nail_care:

reddit_adzerk/adzerkpromote.py
```diff
@@ -464,6 +471,14 @@ def update_flight(link, campaign):
             log_text = None
     else:
         d.update({'Name': campaign._fullname})
+
+        g.ad_events.ad_upsert_request(
+            request_type='create',
+            ad_object_type='flight',
+            ad_object_fullname=campaign._fullname,
+            request_payload=json.dumps(d),
```
drunken-economist: This will be easier to work with in querying (interana, hive, etc) if you don't nest all the relevant information under another `payload.request_payload.*` level. Bump them all up to fields directly in the payload

reddit_adzerk/lib/events.py
drunken-economist: ++
Would it be fair to consider the ad object the "target" of this event? If so, I'd suggest `target_fullname`, `target_type`, `target_id`, etc

drunken-economist:
:nail_care: Agreed with @dwick above, also a suggestion to flatten the payload

zeantsoi:
@dwick agreed that the user initiating the change would be helpful. I've been thinking how to best do this... maybe a _last_updated_by_ attribute on the campaign object?

dwick:
can you not grab it out of `c.user`? most events take a `request` and `context` object to make testing possible.

dwick:
ah, wait this is the adzerk plugin, whoops. let me think on that

zeantsoi:
:haircut: 

Updated event spec [here](https://reddit.atlassian.net/wiki/pages/viewpage.action?pageId=48005166).

reddit_adzerk/adzerkpromote.py
```diff
@@ -476,7 +478,9 @@ def update_flight(link, campaign):
             request_type='create',
             ad_object_type='flight',
             ad_object_fullname=campaign._fullname,
-            request_payload=json.dumps(d),
+            user_id=user._id if user else None,
+            request_payload=d,
+            keywords=keywords,
```
dwick: keywords are already in the payload, why duplicate them?

reddit_adzerk/lib/events.py
```diff
@@ -120,14 +120,16 @@ def ad_upsert_request(
...
         ):
         """
         Create an `ad_create_request` or `ad_update_request` event for
         event-collector.
 
-        ad_object_fullname: The reddit fullname of the ad object
-        ad_object_type: The type of ad object
+        target_fullname: The reddit fullname of the ad object
+        target_type: The type of ad object ()
         request_payload: The JSON serialized payload of the request
```
dwick: user_id/keywords are undocumented

reddit_adzerk/adzerkpromote.py
zeantsoi: Eh... `Keywords` is a `\n` delimited string, but we want to pass the Python list representation in the event. I've reworked this a bit so that rather than passing the list explicitly, I'm just splitting the dict value, then passing the resulting list.

reddit_adzerk/lib/events.py
zeantsoi: Should be fully documented now.

zeantsoi:
:haircut: 

Did the following thing per conversation with @drunken-economist:
- Extract pairs in the JSON body and send them as individual payload members
- Map all IDs to their named representations in Adzerk
- Don't send anything that is `false`
- Send the timestamp, rather than the Adzerk formatted date
- Send entire payload as a string for logging
- Add `user_name`
- Change `target_type` to represent the classname (in this case, `PromoCampaign`)
- Updated [the event spec](https://reddit.atlassian.net/wiki/pages/editpage.action?pageId=48005166)

You can test this out in staging by running some code in shell; here's are some fake campaigns you can update:

    from r2.models import *
    from reddit_adzerk.adzerkpromote import _update_adzerk

    link = Link._byID36('3yv59k')
    account = Account._by_name('zeantsoi')

    _update_adzerk(link, PromoCampaign._byID36('n7oc'), account)
    _update_adzerk(link, PromoCampaign._byID36('n7o4'), account)
    _update_adzerk(link, PromoCampaign._byID36('nb34'), account)

@drunken-economist: **the value we want to exclude from Interana is `adzerk_request`**.

@dwick: if you're keen to deploy, you'll probably want to squash first.

FYI, in testing, the event object seems to be properly formed:

	{
		'topic': 'ad_upsert_events',
		'timestamp': 1459307934096.7578,
		'payload': {
			'priority_id': 75301,
			'user_id': 52,
			'adzerk_request': '{"Name": "t7_b", "IsUnlimited": false, "OptionType": 1, "RateType": 2, "IsFreqCap": null, "GoalType": 1, "StartDate": "/Date(1459142100000)/", "EndDate": "/Date(1459746000000)/", "SiteZoneTargeting": [{"SiteId": 24950, "IsExclude": false}], "CampaignId": 666, "CustomTargeting": "", "Price": 1.5, "PriorityId": 75301, "IsFullSpeed": false, "GeoTargeting": [], "IsDeleted": false, "Keywords": "s.frontpage", "Impressions": 500, "IsActive": false}',
			'is_active': False,
			'price': 1.5,
			'end_ts': '1459746000000',
			'start_ts': '1459142100000',
			'target_type': 'PromoCampaign',
			'campaign_id': 666,
			'impressions': 500,
			'keywords': ['s.frontpage', 'pics'],
			'is_deleted': False,
			'rate_type': 'CPM',
			'user_name': 'zeantsoi',
			'goal_type': 'Impressions',
			'target_fullname': 't7_b'
		},
		'event_type': 'ss.ad_create_request',
		'id': UUID('a1a68c4b-5041-4152-bf09-a57e29d79b57')
	}

dwick:
closing in favor of https://github.com/reddit/reddit-plugin-adzerk/pull/161

PR #159, opened 2016-03-29
Update all flights when a links `over_18` flag is changed

dwick:
re: https://reddit.atlassian.net/browse/ADS-417

:eyeglasses: @tsegaran


tsegaran:
:fish:

PR #162, opened 2016-03-31
Adzerk api update

tsegaran:
Updates Adzerk API to support active flights and flights for campaings

:eyeglasses: @dwick


reddit_adzerk/adzerk_api.py
```diff
@@ -107,9 +107,9 @@ def _to_data(self):
         return {self._name: json.dumps(self._to_item())}
 
     @classmethod
-    def list(cls):
+    def list(cls, params = {}):
```
zeantsoi: Quick comment: mutable default args are a [common gotcha in Python](http://docs.python-guide.org/en/latest/writing/gotchas/).

tsegaran: Understood, but I'm not modifying it in that function... would you be more comfortable with **params ?

tsegaran: Actually I'll just change it to None, that's what the requests library does

zeantsoi: :thumbs_up:

reddit_adzerk/adzerk_api.py
```diff
@@ -107,9 +107,9 @@ def _to_data(self):
         return {self._name: json.dumps(self._to_item())}
 
     @classmethod
-    def list(cls):
+    def list(cls, params = None):
```
dwick: nit: no spaces between `=` in kwargs

reddit_adzerk/adzerk_api.py
```diff
@@ -281,6 +281,10 @@ def _from_item(cls, item):
                              for item in thing.CreativeMaps]
         return thing
 
+    @classmethod
+    def list(cls, is_active = False):
```
dwick: nit : spaces

reddit_adzerk/adzerk_api.py
```diff
@@ -399,10 +403,10 @@ class Channel(Base):
     _name = 'channel'
     _fields = FieldSet(
         Field('Title'),
-        Field('Commission'), 
-        Field('Engine'), 
-        Field('Keywords'), 
-        Field('CPM'), 
+        Field('Commission'),
```
dwick: can this be a separate commit since it's an unrelated whitespace change?

reddit_adzerk/adzerk_api.py
```diff
@@ -441,6 +445,15 @@ class Campaign(Base):
         Field('Price'),
     )
 
+    def fill_flights(self):
+        url = '/'.join([Campaign._base_url, self._name, str(self.Id), 'flight'])
+        print(url)
```
dwick: `print`

reddit_adzerk/adzerk_api.py
```diff
@@ -441,6 +445,15 @@ class Campaign(Base):
         Field('Price'),
     )
 
+    def fill_flights(self):
```
dwick: what does this do exactly? i'm not following from the name and it's not used yet.  forgive me! :)

reddit_adzerk/adzerk_api.py
dwick: @zeantsoi why are you here?!?! go have fun!

dwick:
:nail_care: is this to be used somewhere else? the methods you've added don't have any invocations yet.

dwick:
:fish: lgtm

PR #163, opened 2016-03-31
Script that polls Adzerk for active flights

tsegaran:
…st in memcache.


tsegaran:
:eyeglasses: dwick

kjoconnor:
How big would we expect these cache objects to be, and how frequently will they be accessed?

tsegaran:
I would expect them to be <100K, accessed on pretty much every page load.

kjoconnor:
OK - is this a proof of concept thing to only run for a short period of time or something that will be permanent?

tsegaran:
I expect it will be permanent, it's necessary for keyword targeting. Is
that going to cause a problem?

dwick:
commit message is too long and you're missing the upstart config (example: https://github.com/reddit/reddit-plugin-adzerk/blob/master/upstart/reddit-job-queue_promo_reports.conf).  looks good otherwise :nail_care:

tsegaran:
:eyeglasses: @dwick

dwick:
:fish: be sure the shorten the comment message when you rebase!

tsegaran:
I don't actually need the subreddit target list in there, it was just
useful for debugging. The system list is small. So it will just be the
keyword list, which will eventually have a few thousand words at ~10 bytes
each.

Also there will be no keywords initially, it will grow over time.

Does that sound more reasonable?

PR #164, opened 2016-04-04
Keyword targeting using Zookeeper

tsegaran:
Periodically polls adzerk for keywords being used for targeting. Adds any changes to Zookeeper.

:eyeglasses: @spladug @dwick


reddit_adzerk/adzerkkeywords.py
```diff
@@ -0,0 +1,37 @@
...
+    for flight in active_flights:
+        for keyword_list in flight.Keywords.split('\n'):
+            for keyword in keyword_list.split(','):
+                ks = keyword.strip()
+                if ks.startswith('k.') or ks.startswith('!k.'):
+                    keyword_target.add(ks)
+
+    # Store results in zookeeper
+    if g.zookeeper:
+        kwl = LiveList(g.zookeeper, "/keyword-targets")
```
spladug: I think I gave you a bum steer with using `LiveList` for this. The problem is that it works by using child nodes and a children watcher. So adding and removing a bunch of nodes will result in a series of updates being broadcast to watching clients. If the delta between keywords is large, this could potentially hurt things. Since we're changing everything at once, it'd be simpler to do a single atomic change in ZK so there's only one update sent out.

You could use a `LiveConfig` for this or just directly do `g.zookeeper.set("/keyword-targets", json.dumps(my_big_list_of_keywords)`.

reddit_adzerk/adzerkkeywords.py
```diff
@@ -0,0 +1,37 @@
+# Polls Adzerk for current list of flights and saves the current targeting
+# information to zookeeper (to be run periodically with upstart)
+
+import adzerk_api
+from pylons import app_globals as g
+try:
+    from r2.lib.zookeeper import LiveList
+except ImportError:
```
spladug: when does this happen?

spladug:
:nail_care: question about live lists. i think using zk is going to work in general though. looks good, thanks!

reddit_adzerk/adzerkkeywords.py
tsegaran: I actually liked the idea of Livelist, since I assume this will change pretty slowly. I can't imagine deltas of more than a few dozen at once. However, I'm happy to change this to whatever you prefer.

reddit_adzerk/adzerkkeywords.py
tsegaran: I used the same try: block that was used elsewhere. I assume this happens when running locally when zookeeper isn't set up.

spladug: ah, i see it in ipneuter. that's probably just legacy, i don't think you need to copy that.

reddit_adzerk/adzerkkeywords.py
spladug: The only things actually watching these lists will be the queue processors that add keywords to link objects right? That might be OK. It does seem a bit wasteful to have independent update notifications when it's really just one big one, but not the end of the world if it's limited.

tsegaran:
Ok, switched it to just a JSON field.

Also, there was a very small change to adzerk_api required for this that was rolled back, so it's included here too.

tsegaran:
:haircut:

spladug:
:fish: looks good to me. thanks!

tsegaran:
746d08c

PR #175, opened 2016-04-21
Dump request and response body to logs

dwick:
👓 @tsegaran @spladug 

Having trouble reproducing and issue with the bourne roadblock on staging so we need some more info.  Not seeing anything weird in any of the statsd graphs either.  @spladug is volume going to be an issue here?  We can probably revert this after we figure out what's going on.

### DIFF
--- a/reddit_adzerk/adzerkpromote.py
+++ b/reddit_adzerk/adzerkpromote.py
@@ -926,13 +926,26 @@ def adzerk_request(keywords, uid, num_placements=1, timeout=1.5,
     finally:
         timer.stop()
 
+    errored = False
+
     try:
         response = adzerk_api.handle_response(r)
     except adzerk_api.AdzerkError:
         g.stats.simple_event('adzerk.request.badresponse')
         g.log.error('adzerk_request: bad response (%s) %r', r.status_code,
                     r.content)
-        return None
+        errored = True
+    finally:
+        # Temporarily log request data and response body,
+        # sample at 1%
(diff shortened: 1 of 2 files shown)


tsegaran:
I think we'll need to sample so as not to overwhelm the log server. 1% should be enough to diagnose frontpage problems...
💅

dwick:
💇  https://github.com/reddit/reddit-private/pull/1396

reddit_adzerk/adzerkpromote.py
```diff
@@ -928,6 +928,12 @@ def adzerk_request(keywords, uid, num_placements=1, timeout=1.5,
...
         response = adzerk_api.handle_response(r)
+
+        # Temporarily log request data and response body,
+        # sample at 1%
+        if random.random() < g.live_config.get('ad_log_sample_rate', 0):
+            g.log.info("ad_request: %s, ad_response: %s", json.dumps(data), json.dumps(r))
+
     except adzerk_api.AdzerkError:
         g.stats.simple_event('adzerk.request.badresponse')
         g.log.error('adzerk_request: bad response (%s) %r', r.status_code,
```
tsegaran: Add the request here too, so we see it for failures as well.

tsegaran:
💅

dwick:
💇

spladug:
:fish: but note that you might run into syslog limits with those json responses. individual log entries can only be so large. (8k i think?)

dwick:
trying something else

PR #178, opened 2016-04-27
Rename `link_id` and `campaign_id` in ad_response events.

dwick:
These are actually fullnames so the suffix should be `_fullname`.

👓 @jophuds


jophuds:
🐟

PR #185, opened 2016-05-06
Add geoip_region to ad serving events

dwick:
👓 @tsegaran


dwick:
we can use this data to get reasonable estimates of audience size by state

reddit_adzerk/lib/events.py
```diff
@@ -20,6 +23,10 @@ def get_context_data(cls, request, context):
         if dnt_header is not None:
             data["dnt"] = dnt_header == "1"
 
+        location = location_by_ips(request.ip)
```
zeantsoi: At this point in time, we can only target regions within the US. Do we want to forgo the lookup if we know the request is coming from outside the US?

zeantsoi:
Is this already being tracked by data on pageview requests? If so, we can use that data as a proxy for ad requests.

spladug:
hey, events have a `geoip_country` field already which is fetched from more efficient sources (cdn-provided headers) rather than querying in-request the geoip service. additionally, denormalized data like this should soon not be sent at all from the server and rather computed in midas (the enrichment layer). i don't think this is really a good/right approach to get this data into your events.

Kaitaan:
I actually just added the midas module to ad_serving_events (Midas went live today).  You should see fields for client_geo_country, client_geo_city, client_geo_region, and client_geo_dma_code in hive and interana now.  Let me know if you have any questions or see any weirdness.

dwick:
thanks @Kaitaan 

https://interana.data.uw2.snooguts.net/s/q89n10U

PR #186, opened 2016-05-10
Use unique clicks for click counts

dwick:
👓 @zeantsoi 

docs: http://help.adzerk.com/hc/en-us/articles/208580316-Click-Bucketing
ticket: https://reddit.atlassian.net/browse/ADS-458

No plans to backfill since I think that will just cause confusion.


zeantsoi:
🐟

PR #199, opened 2016-06-14
Add an impression goal to cpm auction campaigns

dwick:
Since caps have no pacing add the theoretical maximum number of
impressions as the impression goal to prevent things from wildly
overspending.

👓 @tsegaran @zeantsoi


tsegaran:
Note that this will show, self-serve campaigns as almost always being "in danger"... the daily and frequency caps will still apply right?

tsegaran:
If you can check that this won't mess up the pacing otherwise, this is a 🐟

zeantsoi:
Won't this also likely result in (potentially considerable) overdelivery? [This calculation](https://github.com/reddit/reddit-plugin-adzerk/pull/199/files#diff-8ff7bdcffc2cdf19a06e2d539306378aR478) will almost certainly result in an impression count that is higher than that what the campaign would otherwise deliver. If we're not throttling on spent budget, we'll be delivering as though the flight is winning on a minimum CPM basis where we actually want to deliver in accordance to the submitted bid, no?

zeantsoi:
Are budget caps and impression goals mutually exclusive? _Goal type_ is seemingly fixed to _percentage_ in the Adzerk UI for auction campaigns:

screen shot 2016-06-14 at 6 11 57 pm

zeantsoi:
💅

dwick:
@zeantsoi Caps have no throttling effect, they're simply "maximums". I would think adzerk would use the more restrictive of the two (between the cap and the goal), but I will confirm this.

@tsegaran Frequency caps still apply here.  Also I'm not sure about the "in danger" thing with respect to having a goal + a cap, but I don't think it would matter since we don't do anything in response to those states.

tsegaran:
The general idea definitely makes sense, I think it makes sense to check in with Adzerk to see if this will do what we expect (slow delivery of people trying to break our caps)

dwick:
:+1: asked in slack, will resubmit for review when i hear back.

dwick:
screen shot 2016-06-14 at 7 15 01 pm

💇

zeantsoi:
:fish: not at my computer ATM: when campaigns go live, are they reupdated in Adzerk with the latest attributes in Reddit? If not, we may want to flush existing campaigns through update_promos or whatever the sync function is.

dwick:
yeah, i'll check that out and refresh accordingly

PR #201, opened 2016-06-16
Add an impression goal to cpm auction campaigns

dwick:
Since caps have no pacing add the theoretical maximum number of
impressions as the impression goal to prevent things from wildly
overspending.

👓 @zeantsoi @tsegaran 

this was already fished, looks like z hit the merge button by accident though: https://github.com/reddit/reddit-plugin-adzerk/pull/199


zeantsoi:
🐟 

https://github.com/reddit/reddit-plugin-adzerk/pull/199 was force pushed over so this code is not actually committed yet.

PR #198, opened 2016-06-14
Check needs_review when activating in flight in Adzerk

zeantsoi:
👓 @dwick 

Per [reddit-public #5381](https://github.com/reddit/reddit-public/pull/5381).

PR #203, opened 2016-06-20
Check campaign_needs_review when activating in flight in Adzerk

zeantsoi:
👓 @dwick 

Reopens https://github.com/reddit/reddit-plugin-adzerk/pull/198:

Having some weirdness pushing from local to the respective branch on GitHub... looks like it may have gotten detached somehow. This reopens the PR.


reddit_adzerk/adzerkpromote.py
```diff
@@ -417,7 +417,10 @@ def update_flight(link, campaign, triggered_by=None):
     elif targets_frontpage:
         keywords.add("s.frontpage")
 
-    campaign_needs_approval = campaign.needs_approval
+    campaign_needs_approval = True
+    if (not promote.campaign_needs_review(campaign, link) and
```
dwick: would this be easier to understand without the `if`?

```python
campaign_needs_approval = (promote.campaign_needs_review(campaign, link) and
                                                not campaign.is_approved)
```

dwick:
💅

reddit_adzerk/adzerkpromote.py
zeantsoi: Yep, made the change (had to switch the order of evaluation to conform to PEP-8 line length requirements).

I've updated the code to reflect the `promote.campaign_needs_review` method. Regarding [this comment](https://github.com/reddit/reddit-plugin-adzerk/pull/198#discussion_r67741962), I think it's better to explicitly call out both conditions because:

1. Abstracting into a model property might obscure whether the return value results from `campaign.needs_review` or `not campaign.is_approved` (or both)
2. This is the only place in the codebase that'd use the property ATM

zeantsoi:
💇

dwick:
🐟

zeantsoi:
05c3cb2

PR #204, opened 2016-06-21
Use mcrouter/gencache

bsimpson63:
* Change the cache key to use a mcrouter routable prefix
* Use stalecache

👓 @dwick

### DIFF
--- a/reddit_adzerk/lib/cache.py
+++ b/reddit_adzerk/lib/cache.py
@@ -4,17 +4,17 @@ from r2.models import PromoCampaign
 
 class PromoCampaignByFlightIdCache():
     @classmethod
-    def _cachekey(cls, flight_id):
-        return "promo.flight.%d" % flight_id
+    def _cache_key(cls, flight_id):
+        return "flightid:%s" % flight_id
 
     @classmethod
     def add(cls, campaign):
-        cachekey = cls._cachekey(campaign.external_flight_id)
-        g.cache.set(cachekey, campaign._fullname, time=60*60*24)
+        key = cls._cache_key(campaign.external_flight_id)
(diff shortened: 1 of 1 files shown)


reddit_adzerk/lib/cache.py
```diff
@@ -4,22 +4,21 @@
...
 
     @classmethod
     def get(cls, flight_id):
-        fullname = g.cache.get(cls._cachekey(flight_id))
+        fullname = g.gencache.get(cls._cache_key(flight_id), stale=True)
 
         if not fullname:
             q = PromoCampaign._query(
                 PromoCampaign.c.external_flight_id == flight_id,
-                data=True,
```
dwick: isn't `data=True` required when querying on attributes not in `Thing`?

dwick:
💅 one question

reddit_adzerk/lib/cache.py
bsimpson63: Not anymore. We now always load the data attributes and `data=True` is meaningless.

bsimpson63:
💇

dwick:
🐟

reddit_adzerk/lib/cache.py
dwick: 🔪 🔪 🔪 🔪 🔪 🔪

zeantsoi: > We now always load the data attributes and `data=True` is meaningless.

@bsimpson63, is this true of everything that subclasses `Thing`?

bsimpson63: Yeah, `data=True` is meaningless everywhere and can be removed.

bsimpson63:
merged

PR #209, opened 2016-07-14
PromoCampaignByFlightIdCache: Ignore errors on cache set

bsimpson63:
Hopefully we can get the caches to fail less often, but in the meantime this will cut down on some errors that can be ignored.

👓 @dwick @MelissaCole


MelissaCole:
:fish:

bsimpson63:
merged

PR #212, opened 2016-07-19
Do not access g.az_selfserve_priorities in module scope

zeantsoi:
👓 @bsimpson63 

Quick fix per [reddit-public #5539](https://github.com/reddit/reddit-public/pull/5539):

If `private` isn't installed, `g.az_selfserve_priorities` isn't set and throws an error when the module is loaded. This moves the accessing of the config var into the function scope, which really shouldn't matter since this is the only place where the config var is ever read.


bsimpson63:
🐟 nice, I think I've been bit by this before on tests or something

zeantsoi:
c21523110369449f3a3b6bcdd4ed795e5c9f4df3

PR #213, opened 2016-07-20
update loid_created usage to be compat with new datetime format

KeyserSosa:
I'm updating how these are stored in https://github.com/reddit/reddit-public/pull/5549, which is currently partially deployed.  The datetime changes are not currently compatible with this plugin though. 

👓  @zeantsoi @dwick


dwick:
🐟 thanks for catching this

PR #215, opened 2016-07-21
Add interests to ad properties for programmatic

tsegaran:
👓 @dwick


reddit_adzerk/adzerkads.py
```diff
@@ -37,6 +37,10 @@ def __init__(self, displayed_things=[], link=None):
         if link:
             properties["page"] = link._fullname
 
+        interests = [w[2:] for w in keywords if w.startswith('a.')]
+        if len(interests) > 0:
+            properties["interests"] = ','.join(interests)
```
dwick: custom properties can be arrays:

http://help.adzerk.com/hc/en-us/articles/203194245-Setting-Custom-Targeting-via-the-API
http://help.adzerk.com/hc/en-us/articles/203194385-Zerkel-Queries#Operators_for_Array_Values

you sure you want to make this a string?

dwick:
💅 just a question

tsegaran:
See the first comment here: https://reddit.atlassian.net/browse/ADS-613

We have to be able to pass it as a custom URL parameter to the exchange. I'm not sure what will happen if I put an array instead of a list in that template language. The partner confirmed that a comma-separated list would work. 

I agree that an array would be cleaner but I'm not sure about the behavior on Adzerk's side.

💇

tsegaran:
Ok, it looks like it's not passing these in as javascript properties but as a template-able string, so it'll be easier to keep it as a string. 💇

jamesmaa:
🐟 LGTM

tsegaran:
After slack discussion, this is now an array 💇

dwick:
🐟

PR #216, opened 2016-07-21
Add cache busting string to ados.js

dwick:
👓 @aoiwelle 

Adzerk made an update to ados.js to fix the adx passback, but the script is cached with a 7 day TTL.  This ensures we have the latest version of ados via a cache busting param tied to r2's short hash.


dwick:
another option would be to make this url configured via live_config so we can bust the cache when we want to, but won't otherwise.  thoughts?

aoiwelle:
This seems su-:fish:-ent.

PR #217, opened 2016-07-25
Search for the correct window before firing the passback

dwick:
👓 @aoiwelle 

This crazy bit of code looks for the `divName` property in windows above the adx passback.  If it errors it has hit an unfriendly iframe so it fires off `postMessage` events to every frame at that level, which hopefully is the correct frame.

Seemed to work when tested via the js console.  Current code doesn't work so I don't see any harm in trying this.


aoiwelle:
🐟

PR #218, opened 2016-07-26
Stop showing CPC ad after a user clicks on it

dwick:
👓 @tsegaran 

ticket: https://reddit.atlassian.net/browse/ADS-653


tsegaran:
That was quick!

🐟

PR #219, opened 2016-07-27
Add frame class for double sidebar feature

jamesmaa:
JIRA: https://reddit.atlassian.net/browse/ADS-631
Other PR: https://github.com/reddit/reddit-public/pull/5605
👓 : @dwick


reddit_adzerk/adzerkads.py
```diff
@@ -59,6 +59,7 @@ def __init__(self, displayed_things=[], link=None):
 
         self.ad_url = g.adzerk_url.format(data=json.dumps(data))
         self.frame_id = "ad_main"
```
dwick: the id won't be unique since we call this twice with the same id: https://github.com/reddit/reddit-public/pull/5605/files#diff-4fb6e010c4a59a377bc221c21cda4025L28

dwick: note that there is JS that relies on these ids:

https://github.com/reddit/reddit-plugin-adzerk/blob/7aa078db1217d8b28fb7218dfec858ecc57f1b0d/reddit_adzerk/public/static/js/adzerk/companion.js#L6

https://github.com/reddit/reddit-plugin-adzerk/blob/7aa078db1217d8b28fb7218dfec858ecc57f1b0d/reddit_adzerk/public/static/js/adzerk/adzerk.js#L43

https://github.com/reddit/reddit-plugin-adzerk/blob/7aa078db1217d8b28fb7218dfec858ecc57f1b0d/reddit_adzerk/templates/baseadframe.html#L55-L59

dwick:
💅

jamesmaa:
Update: I'm been trying to untangle all the assumptions made in the code surrounding the `ad_main` id. 

I decided to resolve the issue by creating a new ad type with frame_id `ad_main_top`.

Does that solution feel particularly roundabout? 
💇

jamesmaa:
Is 
https://github.com/reddit/reddit-plugin-adzerk/blob/7aa078db1217d8b28fb7218dfec858ecc57f1b0d/reddit_adzerk/public/static/js/adzerk/adzerk.js#L43
Something I have to worry about? It seems to be sending Moat data, but I'm not sure if I should replicate one for `ad_main_top` since it's sending the same data

https://github.com/reddit/reddit-plugin-adzerk/blob/7aa078db1217d8b28fb7218dfec858ecc57f1b0d/reddit_adzerk/public/static/js/adzerk/companion.js#L6

Same thing with this ^^, since this is the sponsorship box, which won't load in the experiment

dwick:
those 2 are only for the 300x100, so i think you're ok there.

dwick:
this is going to double the number of banner requests to adzerk.  previously we requested both the 300x100 and 300x250 in the same call, but now both iframes make their own calls.  also i think this will break companions (for the above reason). do we know if sales plans to serve both 300x250s as the same ad?

dwick:
💅 discussed the 2 routes in slack as well

tsegaran:
Sales will def be upset if we break the ability to do companions.

dwick:
@tsegaran would now be a good time to roll banners into api/request_promo?

jamesmaa:
💇  This is for a few subreddits. @dwick would you happen to know a good solution around campanions? I don't know the companion logic very well.

dwick:
@jamesmaa basically you need to request ads for both placements at once in 1 request so that the ad server knows that they will be serving together.  the way this worked with the 300x100 is the 300x250 would request both ads and then create the 300x100 iframe.  what you're doing now is having 2 300x250s both request ads individually

jamesmaa:
Could you tell me what `PLACEMENT_TYPES` means in display?

https://github.com/reddit/reddit-plugin-adzerk/blob/master/reddit_adzerk/public/static/js/adzerk/display.js#L8

I'm thinking about adding a new placement type so I can make a separate call from `main`

jamesmaa:
Checking in on this. I'm having a hard time understanding how display.js interacts with ados.js. What does the data flow look like? I tried adding a new `PLACEMENT_TYPES` for the top banner but I'm not getting a response back in `global.ados_ads` for it.

jamesmaa:
chatted w dave on this

jamesmaa:
UPDATE: 
I changed the logic such that the top ad waits for the bottom ad to request to Adzerk before loading its contents, very similar to how the `SponsorshipBox` is loaded. For transparency and education, the dataflow looks like this:
1. `ad_main` and `ad_main_top` are loaded via pages.py. The ads are differentiated using the `frame_id`.
2. In `ads.html`, there is javascript function to load an iframe with /ads/display/300x250#{keywords...properties...}. `ad_main` calls this function first. `ad_main_top` waits (i.e. event handler for message) for `ad_main` to finish calling.
3. The iframe hits an endpoint on our adzerk plugin (found in adzerkads.py), which fetches the 300x250.html template
4. Inside `ad300x250.html`, it uses display.js to request the ads. I changed it so that on top of requesting the sponsorshipbox and the main, it also requests a placement for `top`.
5. Once it's done making calls to Adzerk, it posts a message to the parent window indicating the results are loaded.
6. `ad_main_top`, which has been waiting this time, gets the message, and finally loads its Ad. It also calls the adzerk plugin and loads in `display.js`, but there is a branching logic that tells it to use the existing ados_results instead of calling Adzerk itself. 
7. `ad_main_top` loads its contents.

I borrowed parts of the logic through other parts of the codebase.

Questions:
1. I don't understand `r.hooks` and `r.frames` very well to know exactly what side effects these changes might produce.
2. I want to make sure Adzerk has a way to differentiate the top and bottom ads through reporting and analytics. I added the frame_id to the property, but I'm not sure if it'll surface in the reporting. Any ideas?

Currently on staging:
https://james_maa.staging.snooguts.net/r/gifs/

💇

reddit_adzerk/public/static/js/adzerk/display.js
```diff
@@ -151,6 +181,18 @@
...
           }
         }
+
+        // Load top ad if exists
+        if (global.ados_results.top) {
+          if (global.postMessage) {
+            global.parent.postMessage('ados.createAdFrame:top', config.origin);
+          } else {
+            iframe = document.createElement('iframe');
+            iframe.src = '/static/createadframe.html';
```
dwick: this won't work since it calls `r.adzerk.createSponsorshipAdFrame`. you can get rid of this else case entirely since we don't support browsers this old anymore.

reddit_adzerk/public/static/js/adzerk/display.js
```diff
@@ -98,6 +119,15 @@
 
         request.placement_types.push(PLACEMENT_TYPES[type]);
       }
+    } else if (properties.double_sidebar) {
+      var type = 'main';
```
dwick: you could define `PLACEMENT_TYPES` based on `properties.double_sidebar` and avoid this case

reddit_adzerk/public/static/js/adzerk/display.js
```diff
@@ -34,6 +35,26 @@
   var config = getConfig();
   var properties = config.properties || {};
 
+  // If double sidebar experiment, have the top ad not call Adzerk
+  if (properties.double_sidebar && properties.frame_id == 'ad_main_top') {
```
dwick: 🔕 would it be feasible to have this replace companion.js for the sponsorship box? you could change this if statement to be `properties.frame_id !== 'ad_main'`. not super important.

dwick:
💅 looks good, have an appointment, but i'll look at the other one later tonight

reddit_adzerk/public/static/js/adzerk/display.js
jamesmaa: yeah that should be possible since I ripped it from companion.js anyways. I'll do it in another PR

jamesmaa:
whew I'm glad this approach is suitable. 💇

dwick:
🐟

reddit_adzerk/public/static/js/adzerk/display.js
```diff
@@ -34,6 +35,26 @@
   var config = getConfig();
   var properties = config.properties || {};
 
+  // If double sidebar experiment, have the top ad not call Adzerk
```
dwick: should this be at the top and be the only thing that runs for ad_main_top? i think we're maybe making the ad request in ad_main_top too even though we ignore it?

dwick:
💅 one last question: https://github.com/reddit/reddit-plugin-adzerk/pull/219/files#r74169027

reddit_adzerk/public/static/js/adzerk/display.js
jamesmaa: I'll add a `return` statement at the bottom of the if statement. I put it there because I need to call `getConfig` for the conditional

jamesmaa:
UPDATE:
Adding the return statement made me realize that each ad was still calling its own Adzerk request. After adding the return statement, I found out that the top ad was not showing. After some digging through `ados.js` and the passbacks (not pleasant), I found out that by setting `PLACEMENT` = `top`, ados is looking for a div with id `main`. 

I had to change the div beforehand to load the ad in the proper place. It's kind of hacky but I don't know enough about Mako templates to be able to dynamically set it in `ad300x250.html`.

💇

dwick:
hrm, i think the other option would be to use a query parameter. ados is designed to be put on a single page, not this mess of iframes, hence why we're having so many issues.

💅

dwick:
🐟 after talking on slack

jamesmaa:
b883593cf928e9021c973e692caaae009e379c52

PR #222, opened 2016-08-04
Fix priority id not being mapped to names

dwick:
👓 @aoiwelle 

everything is currently coming in as `unknown (number)` because `priorityId` is a string.

https://interana.data.uw2.snooguts.net/s/b389RL

ticket: https://reddit.atlassian.net/browse/ADS-660


aoiwelle:
🐟

PR #224, opened 2016-08-10
Use specific placement names instead of number `div`s

dwick:
👓 @aoiwelle 

related: https://github.com/reddit/reddit-public/pull/5658


aoiwelle:
🐟

PR #226, opened 2016-08-10
Fix `KeyError` related to looking up ad type friendly names

dwick:
👓 @bsimpson63

### DIFF
--- a/reddit_adzerk/adzerkpromote.py
+++ b/reddit_adzerk/adzerkpromote.py
@@ -1106,7 +1106,7 @@ def adzerk_request(
                 keywords=keywords,
                 platform=platform,
                 placement_name=placement_name,
-                placement_type=AD_TYPE_FRIENDLY_NAMES[ad_type],
+                placement_type="adserver_ad",
                 ad_id=ad_id,
                 impression_id=impression_id,
                 matched_keywords=matched_keywords,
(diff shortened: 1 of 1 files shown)


reddit_adzerk/adzerkpromote.py
```diff
@@ -1135,7 +1135,8 @@ def adzerk_request(
         priority_id = body.get('priorityId', None)
         # default to leaderboard since old creatives will
         # always be a leaderboard, but may not have an ad type defined
-        ad_type = body.get('adType', LEADERBOARD_AD_TYPE)
+        ad_type = int(body.get('adType', LEADERBOARD_AD_TYPE))
+        placement_type = AD_TYPE_FRIENDLY_NAMES.get(ad_type, "unknown type (%d)" % ad_type)
```
bsimpson63: what happens if it's unknown?

bsimpson63: feel free to ignore, but this would be cooler as a `try/except KeyError`

bsimpson63:
🐟 with comments you can ignore

reddit_adzerk/adzerkpromote.py
dwick: this is just to prevent errors if it's not in the `AD_TYPE_FRIENDLY_NAMES`, should only happen if someone forgets to update that dictionary

PR #228, opened 2016-08-12
Set is_banner to True for banners

jamesmaa:
@dwick 👓

### DIFF
--- a/reddit_adzerk/adzerkads.py
+++ b/reddit_adzerk/adzerkads.py
@@ -25,11 +25,13 @@ class Ads(BaseAds):
 
         keywords = promote.keywords_from_context(
             c.user, c.site,
-            frame_id=frame_id,
             include_subscriptions=False,
             displayed_things=displayed_things,
             block_programmatic=getattr(link, 'block_programmatic', False)
         )
+        if promote.ads_feature_enabled("double_sidebar"):
+            keywords.add("exp.double_sidebar.test_group")
+            keywords.add("exp.double_sidebar.ad_type.%s" % frame_id)
 
         properties = adzerkpromote.properties_from_context(
             context=c,
--- a/reddit_adzerk/public/static/js/adzerk/display.js
+++ b/reddit_adzerk/public/static/js/adzerk/display.js
@@ -33,7 +33,6 @@
 
   var config = getConfig();
   var properties = config.properties || {};
-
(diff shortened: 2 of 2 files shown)


reddit_adzerk/adzerkads.py
```diff
@@ -28,7 +28,8 @@ def __init__(self, frame_id="ad_main", displayed_things=[], link=None):
             frame_id=frame_id,
             include_subscriptions=False,
             displayed_things=displayed_things,
-            block_programmatic=getattr(link, 'block_programmatic', False)
+            block_programmatic=getattr(link, 'block_programmatic', False),
+            is_banner=True
```
dwick: i think it might be easier to just add the experiment keywords after this call instead of passing `frame_id` and `is_banner` into the function.  that's the only reason they passed to `keywords_from_context`, right?

```
keywords = promote.keywords_from_context(...)
if promote.ads_feature_enabled("double_sidebar"):
  keywords.add("exp.double_sidebar.test_group")
  keywords.add("exp.double_sidebar.ad_type.%s" % frame_id)
```

dwick:
💅

jamesmaa:
💇

reddit_adzerk/public/static/js/adzerk/display.js
```diff
@@ -131,8 +130,13 @@
     } else {
       for (var type in PLACEMENT_TYPES) {
         placement = ados_add_placement(NETWORK, SITE, type, PLACEMENT_TYPES[type]);
-        placement.setProperties(properties);
-
+        var props;
+        if(properties.double_sidebar && type === 'top'){
+          props = Object.assign({}, properties, {frame_id: 'ad_main_top'});
```
dwick: since you're only setting a single property i would just do that:
```
properties. frame_id = 'ad_main_top';
```

[Object.assign](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign#Browser_compatibility) has limited browser support.

dwick:
[this](https://github.com/reddit/reddit-plugin-adzerk/pull/228/files#diff-0726191060b548e3287886435feead3bR28) can be removed now too.

💅

jamesmaa:
Sonuds good 💇

dwick:
🐟

PR #236, opened 2016-08-18
Encode properties only for sending them to ados

dwick:
👓 @aoiwelle 

modifying the `properties` object was turning all the values to strings before publishing events.  we only need this for setting them w/ados.


aoiwelle:
🐟

PR #238, opened 2016-08-24
Request link reports for date ranges of active campaigns only

dwick:
The issue here is that we were previously looking up every campaign for the links which could potentially include really old ones.  The result was that the start_date of the report would be set so early that the report would take forever / timeout.  we're only interested in current data for these reports since it's a daily breakdown so we should only ask for that.

👓 @tsegaran or @aoiwelle 

Tested this in reddit-shell on my staging w/this: https://gist.github.com/dwick/2db8970032b29db7d81a40e0ebc2cafa

### DIFF
--- a/reddit_adzerk/adzerkreporting.py
+++ b/reddit_adzerk/adzerkreporting.py
@@ -59,28 +59,30 @@ def queue_promo_reports():
     prev_promos = promote.get_served_promos(offset=-1)
     promos = promote.get_served_promos(offset=0)
 
-    links = set()
     campaigns = set()
     campaigns_by_link = defaultdict(set)
 
     for campaign, link in itertools.chain(prev_promos, promos):
-        links.add(link)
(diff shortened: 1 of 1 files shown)


aoiwelle:
🐟

reddit_adzerk/adzerkreporting.py
```diff
@@ -59,28 +59,30 @@ def queue_promo_reports():
...
-    links = set()
     campaigns = set()
     campaigns_by_link = defaultdict(set)
 
     for campaign, link in itertools.chain(prev_promos, promos):
-        links.add(link)
         campaigns.add(campaign)
         campaigns_by_link[link].add(campaign)
 
+    links = [(link, campaigns) for link, campaigns in campaigns_by_link.items()]
```
aoiwelle: 🔕  You could probably use a generator here, especially since you're only using the sorted version later.

PR #239, opened 2016-08-24
Remove companion

jamesmaa:
👓 : @dwick


reddit_adzerk/adzerkads.py
```diff
@@ -34,9 +34,7 @@ def __init__(self, frame_id="ad_main", displayed_things=[], link=None):
             displayed_things=displayed_things,
             block_programmatic=getattr(link, 'block_programmatic', False)
         )
-        if promote.ads_feature_enabled("double_sidebar"):
-            keywords.add("exp.double_sidebar.test_group")
-            keywords.add("exp.double_sidebar.ad_type.%s" % frame_id)
+        keywords.add("properties.frame_id.%s" % frame_id)
```
dwick: is this supposed to be a property?  there's an array of them below.  otherwise i'd go with something like `s.placement.%s % frame_id`

reddit_adzerk/public/static/js/adzerk/display.js
```diff
@@ -5,7 +5,7 @@
 
   var PLACEMENT_TYPES_FRIENDLY_NAMES = {
     '5': '300x250',
-    '8': '300x100',
+    '8': '300x100', // deprecated ad type
```
dwick: do we actually need to keep this?

reddit_adzerk/public/static/js/adzerk/display.js
```diff
@@ -21,9 +21,10 @@
   }, {});
   var PLACEMENT_TYPES = {
     main: 5,
-    sponsorship: 8,
+    top: 5,
```
dwick: does it make sense to rename `main` -> `bottom` while we're at it?

dwick:
💅 nothing major

PR #146, opened 2016-02-04
Aggregate reporting records by campaign/date

dwick:
Adzerk responds with multiple details records if the price changes.
We need to aggregate on those.

From them:
> If you change the rate of a flight in the middle of the day (or in between start adn end date as it was the case here), we'll show the impressions for the one rate in one line item, and the impressions for the other rate in another.

:eyeglasses: @zeantsoi

PR #241, opened 2016-08-31
Fix campaign reports saving incorrect results when price changes

dwick:
👓 @zeantsoi 

same as this issue: https://github.com/reddit/reddit-plugin-adzerk/pull/146 but for the lifetime reports instead of the daily ones.

### DIFF
--- a/reddit_adzerk/adzerkreporting.py
+++ b/reddit_adzerk/adzerkreporting.py
@@ -335,6 +335,7 @@ def _process_lifetime_campaign_reports(campaigns, report_id, queued_date):
     report_records = report_result.get("Records", None)
     campaigns_by_fullname = {c._fullname: c for c in campaigns}
 
+    campaign_results = defaultdict(_reporting_factory)
     if report_records:
         for detail in report_records[0].get("Details", []):
             campaign_fullname = _get_fullname(PromoCampaign, detail)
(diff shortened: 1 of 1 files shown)


dwick:
issue: https://reddit.atlassian.net/browse/ADS-743

reddit_adzerk/adzerkreporting.py
```diff
@@ -355,13 +356,20 @@ def _process_lifetime_campaign_reports(campaigns, report_id, queued_date):
...
-            campaign.adserver_spent_pennies = int(spent * 100)
-            campaign.adserver_impressions = impressions
-            campaign.adserver_clicks = clicks
-            campaign.last_lifetime_report = report_id
-            campaign.last_lifetime_report_run = queued_date
+            # if the price changes there may be more than 1 record for a single campaign.
+            campaign_values = campaign_results[campaign]
+            campaign_values["impressions"] = campaign_values["impressions"] + impressions
+            campaign_values["clicks"] = campaign_values["clicks"] + clicks
+            campaign_values["spent_pennies"] = campaign_values["spent_pennies"] + (spent * 100.)
```
zeantsoi: What do we do with `campaign_values`?

reddit_adzerk/adzerkreporting.py
```diff
@@ -335,6 +335,7 @@ def _process_lifetime_campaign_reports(campaigns, report_id, queued_date):
     report_records = report_result.get("Records", None)
     campaigns_by_fullname = {c._fullname: c for c in campaigns}
 
+    campaign_results = defaultdict(_reporting_factory)
```
zeantsoi: Does this only need to have the function passed, or does it need to be called as `_reporting_factory()`?

reddit_adzerk/adzerkreporting.py
dwick: https://github.com/reddit/reddit-plugin-adzerk/pull/241/files#diff-a369a8ee1a6bfac28a8b167eb9f58fa7R365

zeantsoi: Oh, I see. Passed-by-reference. 👍

zeantsoi:
[One comment](https://github.com/reddit/reddit-plugin-adzerk/pull/241/files#r77036390) 💅

reddit_adzerk/adzerkreporting.py
dwick: no, when a default dict receives a function it calls that when the key is missing upon index.

basically:
def defaultdict(fn):
   dict = {}
   return lambda key: dict[key] if key in dict else fn()

dwick:
💇

zeantsoi:
🐟

PR #242, opened 2016-09-08
Pass zones for above_the_fold and below_the_fold banner requests

zeantsoi:
👓 @jamesmaa 

Per [JIRA ADS-726](https://reddit.atlassian.net/browse/ADS-726):

Enables the companioning of two 300x250 banner ads by specifying a "zone" when making the ad request. I've created an [_above the fold_ and _below the fold_](http://reddit.adzerk.net/publisher/7589/site/24950/zones) under the _desktop_ site in Adzerk. This enables us to companion creatives while dictating which ad appears above the fold, and which appears below the fold. Overall, when a creative is added to a flight, a zone can be specified... if the request is for a specific zone, a zone-specific ad will first be served – if none is found, a non-zone-specific ad matching the request will be served.

I've set up [a flight in Adzerk](http://reddit.adzerk.com/network/5146/brand/151998/campaign/484099/option/2268811) to demonstrate how this works. [Here's the corresponding subreddit that displays the ads](https://zeantsoi.staging.snooguts.net/r/MadClownLove).

Note that banner ads continue to be served correctly on other subreddits and listing pages, even though there aren't any zone-specific ads targeted to those pages.


jamesmaa:
I think you need to add zones here too:
https://github.com/reddit/reddit-plugin-adzerk/pull/242/files#diff-a630e5a94de757d8008045ccc1f74b2eR160

Besides that, I'm thinking of other places we would want to instrument this new attribute. Some ideas:
- events and reporting (adzerk_request and adzerk_response)
- impression and click pixels
Anything I might be missing @dwick ? 
Not saying it belongs in this PR, but it's definitely worth considering and doing at some point. I'll write the needed tickets before fishing. 💅

zeantsoi:
> I think you need to add zones here too:
> https://github.com/reddit/reddit-plugin-adzerk/pull/242/files#diff-a630e5a94de757d8008045ccc1f74b2eR160

So we _don't_ actually want to add a zone here. This `ados_add_placement` call is invoked in two situations:

1. [In-feed placements](https://github.com/reddit/reddit-public/blob/master/r2/r2/public/static/js/listing.js#L31)
2. [Spotlight headlines](https://github.com/reddit/reddit-public/blob/master/r2/r2/public/static/js/spotlight.js#L168)

In either scenario, we need to make a call for that specific placement. As of now, neither of those placements have a `zone` and thus we shouldn't be explicitly requesting a zone-specific ad (in contrast to the other calls to `ados_add_placement`, which are _specifically_ requests for above/below-the-fold banners).

💇

reddit_adzerk/public/static/js/adzerk/display.js
```diff
@@ -169,10 +170,18 @@
       }
     } else {
       for (var type in PLACEMENT_TYPES) {
-        placement = ados_add_placement(NETWORK, SITE, type, PLACEMENT_TYPES[type]);
+        var placement;
         if(properties.double_sidebar && type === 'top'){
+          placement = ados_add_placement(NETWORK, SITE, type, PLACEMENT_TYPES[type]);
```
jamesmaa: Why can't this be factored outside of the if-else statement? This seems to be duplicating https://github.com/reddit/reddit-plugin-adzerk/pull/242/files#diff-a630e5a94de757d8008045ccc1f74b2eR181

zeantsoi:
Closing in order to make repo private. Will open up new issue when repo is recreated.

PR #244, opened 2016-09-08
Add user_id to event payload for ad_request

zeantsoi:
👓 @dwick 

Per [JIRA ADS-720](https://reddit.atlassian.net/browse/ADS-720):

Something's not working with downvoting ads. When a user downvotes an ad, it's not supposed to be delivered to that user again. This seems to be broken for some accounts (e.g., it works for _zean_rm_, but not _zeantsoi_). We're also receiving support requests citing this same problem.

I've been debugging with Jacob @ Adzerk and it looks like the downvote pixel is correct, as is the `user` parameter that's being sent in the ad request in staging. However, it's tough to know know what's actually being sent in the production request because `user_id` isn't being sent. This PR sends it with the ad request event.

If this PR is accepted, I'll update the [wiki](https://reddit.atlassian.net/wiki/display/DA/Ad+Serving+Events) with the new attribute.


dwick:
can you add this to responses / anything else that makes sense to have it while you're at it?

adding `user.name` could also be useful for easier usage.

💅

zeantsoi:
💇 

`referrer` and `timeout` seem to be the most interesting attributes that cannot be evinced from elsewhere in the event payload (e.g., `request`, etc.). I've added those here. Any others you can think of?

I'm slightly disinclined to add username since that's really only a heuristic that we (humans) would need. We can always get the requesting user from `request`... `user_id` in this context could refer to either an loid or an id36, which should (but may not) jive with the request (which is why we want to instrument this). What's most important is we know the precise `user_id` in order to know what we're sending to Adzerk.

Also, I think we could get username added via Midas if we wanted.

dwick:
`referrer` is covered here: https://github.com/reddit/reddit-public/blob/19c0dc389462549f8b206a84d59f530ea7db520b/r2/r2/lib/eventcollector.py#L1931-L1934

looking closer it seems user id/name should already be there also:
https://github.com/reddit/reddit-public/blob/19c0dc389462549f8b206a84d59f530ea7db520b/r2/r2/lib/eventcollector.py#L1910-L1917

dwick:
💅 https://interana.data.uw2.snooguts.net/s/35066o

These fields are already there

zeantsoi:
I mean user_id as in the _id36/loid since that's what's being communicated to Adzerk. Need to see the exact value that's being passed in the request, we shouldn't assume that it _necessarily_ corresponds to the hex-converted id from the base request.

Unless you can think of a better way of logging it?

zeantsoi:
Closing in order to delete/recreate master repo, I'll reopen there.

